PR #2, opened 2015-10-16
Remove definition of sonar.analysisMode

henryju:


dbmeneses:
LGTM (I don't know that property) - build failing though

PR #4, opened 2015-10-27
Update spock dependency

dbmeneses:
Otherwise I get the error: "exception org.spockframework.util.IncompatibleGroovyVersionException: The Spock compiler plugin cannot execute because Spock 1.0.0-groovy-2.3 is not compatible with Groovy 2.4.4"


henryju:
You should always try to use the Gradle wrapper to build Gradle projects so that you are sure to use the correct Gradle version. Instead of running `gradle build` just try `./gradlew build`

PR #10, opened 2016-08-11
SONARGRADL-16 removing dependency to internal gradle class

bfarka:
this PR fixed SONARGRADLE-16 it removes to class to the internal JVM class of gradle and backports the piece of code which was used in gradle 2.14.x to the plugin itself.

Since this internal class has changed with gradle 3.x the plugin works with gradle 3.0 too


dbmeneses:
I've rebased it and merged into master. Thanks @bfarka !

PR #13, opened 2016-09-07
Added support for Java FX.

wokis:


henryju:
Hi,

How does Gradle knows how to find the lib? Don't you declare it in your dependencies? Can you share a sample project to reproduce the issue?

Thanks

wokis:
Hi, 

Since Java FX is a part of Java it's automatically added to the class path by the JVM. Java FX is a runtime lib just as rt.jar and it shouldn't be added as a dependency in gradle.

henryju:
According to http://stackoverflow.com/questions/19552740/gradle-fails-to-find-javafx-classes there are several cases: if using Java < 8 then javafx is not part of the JVM classpath. So I guess using the jfx-gradle plugin will add the lib to the project dependencies and so it should works fine with our plugin.
So I guess in your case you are using Oracle JDK 1.8 right?

henryju:
Anyway I have created a ticket:
https://jira.sonarsource.com/browse/SONARGRADL-20

Could you share a sample JavaFX project that I could use in ITs?

wokis:
Correct I'm using JDK 1.8 where the jfx-gradle plugin is no longer needed. I've created a sample hello world project https://github.com/wokis/jfx-sample-project.

henryju:
Thanks! Merged.

PR #20, opened 2016-11-14
Fixed 'sonar.libraries' property population

pscosta:
Fixed 'sonar.libraries' population: library paths were incorrectly being added as a single property, resulting in exceptions during the sonar run as the following:

```
Caused by: java.lang.IllegalStateException: No files nor directories matching '[/path/to/home/.gradle/caches/modules-2/files-2.1/com.example/some-library/1.0/7a885b1624285ea1a7b91fbd31040554a6385a10/some-library-1.0.jar' in directory /path/to/home/my-project
    at org.sonar.batch.scan.ProjectReactorBuilder.validateDirectories(ProjectReactorBuilder.java:302)
```

### DIFF
--- a/src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
+++ b/src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
@@ -299,7 +299,7 @@ public class SonarQubePlugin implements Plugin<Project> {
         appendProp(properties, "sonar.groovy.binaries", mainClassDir);
       }
       // Populate deprecated properties for backward compatibility
-      appendProp(properties, "sonar.binaries", mainClassDir);
+      appendProps(properties, "sonar.binaries", mainClassDir);
     }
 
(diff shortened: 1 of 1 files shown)


pscosta:
Also, the problem was first reported here: http://stackoverflow.com/questions/40499538/sonar-gradle-error-on-invalid-value-of-sonar-libraries/40599308#40599308

pscosta:
Also, the problem was first reported here: http://stackoverflow.com/questions/40499538/sonar-gradle-error-on-invalid-value-of-sonar-libraries/40599308#40599308

henryju:
Nice catch! We usually don't support SonarQube versions older than LTS (current is 5.6) but since you submitted this P/R, I will release a bug fix. Here is a ticket: https://jira.sonarsource.com/browse/SONARGRADL-30

pscosta:
Thanks! v.2.2.1 works wonderfully with my setup now.

PR #23, opened 2017-01-24
Adding an OutputFile to SonarQubeTask.

wg-tsuereth:
The file is a dummy, merely indicating that the task has been run.  But its presence allows an incremental gradle build to recognize when sonarqube's inputs haven't been changed, and it doesn't need to be re-executed.

### DIFF
--- a/src/main/java/org/sonarqube/gradle/SonarQubeTask.java
+++ b/src/main/java/org/sonarqube/gradle/SonarQubeTask.java
@@ -19,6 +19,7 @@
  */
 package org.sonarqube.gradle;
 
+import java.io.File;
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Properties;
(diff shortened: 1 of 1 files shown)


henryju:
SonarQube analysis is depending on many parameters (like server side configuration). I fail to see how this could make incremental analysis properly work?

wg-tsuereth:
That's true.  For my team's purposes, we don't expect any out-of-band configuration to change from build to build.

What I would really have liked to do is implement an option to toggle the output file setting - i.e. enable hacky-incremental sonar runs with "-PSonarIncrementalExperimental=1" - but I'm not enough of a gradle wizard to know how to make the OutputFile conditional.

Meanwhile, this patch is more for my own purposes than a general solution.

henryju:
I'm tempted to create a more generic ticket in our bug tracker about supporting (or not) incremental build. We could then work on a proper specification of the use cases.

henryju:
I'm closing this ticket, since for now we have no plan for supporting incremental analysis. Our analyzer are becoming smarter and need to analyze all files to find control flow issues, even if part of them haven't changed.

PR #29, opened 2017-04-06
SONARGRADL-33 Do not set sonar.junit.reportsPath if there are no test reports to avoid WARN or ERROR messages

Vampire:


bellingard:
Thanks @Vampire for the PR! I've set this to be part of the next release, we'll look into it when we start the sprint on this version.

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -55,13 +55,16 @@
...
+import static java.util.Arrays.asList;
+
 /**
  * A plugin for analyzing projects with the <a href="http://redirect.sonarsource.com/doc/analyzing-with-sq-gradle.html">SonarQube Runner</a>.
  * When applied to a project, both the project itself and its subprojects will be analyzed (in a single run).
  * Please see the “SonarQube Runner Plugin” chapter of the Gradle User Guide for more information.
  */
 public class SonarQubePlugin implements Plugin<Project> {
 
+  private static final Pattern TEST_RESULT_FILE_PATTERN = Pattern.compile("TESTS?-.*\\.xml");
```
henryju: I'm a bit uncertain about hardcoding this. How stable it is?

Vampire: Depends on you guys. See here: https://github.com/SonarSource/sonar-java/blob/master/java-surefire/src/main/java/org/sonar/plugins/surefire/SurefireJavaParser.java#L76-L83 that is exactly what the pattern expresses.

Vampire: And on the Gradle guys. See here: https://github.com/gradle/gradle/blob/master/subprojects/testing-jvm/src/main/java/org/gradle/api/internal/tasks/testing/junit/result/Binary2JUnitXmlReportGenerator.java#L69-L71

Vampire: But if Gradle would produce a different pattern, the SurefireJavaParser wouldn't catch them anyway currently as only the pattern here is considered.

henryju: Alright, let's assume this is a standard then :)

henryju:
2.4-rc1 deployed on the Gradle plugin portal.
https://groups.google.com/forum/#!topic/sonarqube/nVUoJ8juocw

PR #33, opened 2017-10-12
Support Android Gradle 3

Shusshu:
https://jira.sonarsource.com/browse/SONARGRADL-41

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -4,7 +4,7 @@
 plugins {
   id "com.gradle.plugin-publish" version "0.9.6"
   // Eat your own dog food :)
-  id "org.sonarqube" version "2.3"
+  id "org.sonarqube" version "2.5"
   id "com.jfrog.artifactory" version "4.4.4"
   id "com.github.hierynomus.license" version "0.13.1"
   id "net.researchgate.release" version "2.4.1"
@@ -41,6 +41,9 @@ repositories {
   mavenLocal()
   mavenCentral()
   maven {
+    url "https://maven.google.com"
+  }
+  maven {
     url "https://repox.sonarsource.com/sonarsource"
   }
 }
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -41,14 +41,17 @@ repositories {
   mavenLocal()
   mavenCentral()
   maven {
+    url "https://maven.google.com"
```
CedrickFlocon: can be replace by google() ?

Shusshu: I know, it depends which version you use and what dependencies. sometimes it's not an option to use google() so this can be changed in a future PR

kenyee:
Please merge so we can pull snapshot builds if this doesn't break backwards compatibility w/ AGP 2.3.3

src/main/java/org/sonarqube/gradle/AndroidUtils.java
```diff
@@ -211,7 +211,7 @@ private static void populateSonarQubeProps(Map<String, Object> properties, List<
     // I don't know what is best: ApkVariant::getCompileLibraries() or BaseVariant::getJavaCompile()::getClasspath()
```
agacemi: In this case, comments are deprecated

Shusshu:
If you really need this now you can always use my branch and build it yourself or use jitpack
See my answer here: https://stackoverflow.com/a/46813528/1228221

henryju:
Squashed and fixed integration tests in #34. Thanks for your contribution!

jordanjoz1:
@Shusshu it looks like deleting your branch removed the archive on jitpack. Would you mind keeping that branch around until the official update comes out?

henryju:
I have deployed a RC1: https://plugins.gradle.org/plugin/org.sonarqube/2.6-rc1
Feedback is welcome.

ngtignacio:
2.6-rc1 seems to work ok for me, thanks 👍

henryju:
Thanks for the feedback, 2.6 is already out, but better late than never ;)

PR #36, opened 2017-10-30
SONARGRADL-43 Replace sonar.junit.reportsPath by sonar.junit.reportPaths

henryju:


janos-ss:
Btw I see many more places in the repo with `reportsPath`...

henryju:
@janos-ss That's because I kept old property for backward compatibility. But I forgot to update ITs assertion, now this is done.

PR #40, opened 2018-02-13
SONARGRADL-49 Filter unexisting paths before populating sonar.java.te…

henryju:
…st.binaries

### DIFF
--- a/src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
+++ b/src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
@@ -110,13 +110,7 @@ public class SonarQubePlugin implements Plugin<Project> {
       boolean hasSourceOrTest = configureSourceDirsAndJavaClasspath(project, properties, false);
       if (hasSourceOrTest) {
         configureSourceEncoding(project, properties);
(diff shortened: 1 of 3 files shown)


src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -223,8 +223,12 @@ static void appendProp(Map<String, Object> properties, String key, Object valueT
...
 
   static void setTestClasspathProps(Map<String, Object> properties, Collection<File> testClassDirs, Collection<File> testLibraries) {
-    appendProps(properties, "sonar.java.test.binaries", testClassDirs);
-    appendProps(properties, "sonar.java.test.libraries", testLibraries);
+    appendProps(properties, "sonar.java.test.binaries", exists(testClassDirs));
+    appendProps(properties, "sonar.java.test.libraries", exists(testLibraries));
+  }
+
+  private static List<File> exists(Collection<File> files) {
+    return files.stream().filter(File::exists).collect(Collectors.toList());
```
dbmeneses: If this method is only used with directories, perhaps it would be better to use `isDirectory`, which implies that it exists.

henryju: For sonar.java.libraries, it can be files (.jar)

PR #45, opened 2018-05-23
Allprojects

dbmeneses:


src/main/java/org/sonarqube/gradle/SonarQubeTask.java
```diff
@@ -27,11 +27,10 @@
 import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.Properties;
-import org.gradle.api.DefaultTask;
+import org.gradle.api.internal.ConventionTask;
```
henryju: Dependency on internal API?

dbmeneses: good catch

dbmeneses: actually convention mapping is internal. Currently we already import org.gradle.api.internal.ConventionMapping.

PR #47, opened 2018-06-30
Add link to community forum

fitzoh:
Github issues aren't active, so I created a JIRA account which apparently doesn't have permissions to create issues.

alternatively, enable github issues


henryju:
Merged, thanks!

PR #51, opened 2019-02-12
Allow tests to be skipped

AlainODea:
- Conditionally skip adding test tasks to sonarqube task dependencies

This is intended to answer [How do I prevent the Gradle sonarqube task from running tests?](https://community.sonarsource.com/t/how-do-i-prevent-the-gradle-sonarqube-task-from-running-tests/6674/4) on the SonarSource Community.

### DIFF
--- a/src/main/java/org/sonarqube/gradle/SonarQubeExtension.java
+++ b/src/main/java/org/sonarqube/gradle/SonarQubeExtension.java
@@ -59,6 +59,7 @@ public class SonarQubeExtension {
   private boolean skipProject;
   private final ActionBroadcast<SonarQubeProperties> propertiesActions;
   private String androidVariant;
+  private boolean skipTests;
 
   public SonarQubeExtension(ActionBroadcast<SonarQubeProperties> propertiesActions) {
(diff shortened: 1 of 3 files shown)


henryju:
Relates to https://jira.sonarsource.com/browse/SONARGRADL-59. For simplicity, I'm suggesting to completely drop this dependency.

AlainODea:
@henryju that’s even better! 😊

henryju:
Closing this PR, we will drop the dependency on the test task.

mmokeev:
Is there any news about this task? We too need the ability to disable tests in the sonar task.

loganmzz:
Why close PR for "promising" that still has been happened close to 3 years later ....

sgammon:
why is this closed? the `sonarqube` task still requires tests. there are entire project structures this does not work with because the plugin additionally requires installation in the root project, which means you need to run _all_ tests, no matter _what_, for all builds (+ _included builds_) _just_ to get a report from Sonar.

nuhkoca:
This is still happening in 2023. `sonar` task is running all tests itself. Please do something.

PR #53, opened 2019-03-28
Improvement to memory leak concerning SonarQubePlugin#actionBroadcastMap

lkoe:
This PR alleviates a memory leak around `SonarQubePlugin#actionBroadcastMap`.

The problem was reported here: https://community.sonarsource.com/t/memory-leaks-in-sonarqube-gradle-plugin-2-7-and-2-6-2/6165

This PR does not fully fix the issue, but improves upon the current situation by changing the map lookup key to the project name, instead of the full gradle projects.
This enables the gradle daemon to become memory stable again in our development environment.

### DIFF
--- a/src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
+++ b/src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
@@ -67,10 +67,10 @@ public class SonarPropertyComputer {
   static final String SONAR_JAVA_SOURCE_PROP = "sonar.java.source";
   static final String SONAR_JAVA_TARGET_PROP = "sonar.java.target";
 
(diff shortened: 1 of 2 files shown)


dbmeneses:
Thanks for the PR. It would be better to use project's path as key instead of the name. According to javadoc: `The project's name is not necessarily unique within a project hierarchy. You should use the getPath() method for a unique identifier for the project.`
Could you please update the PR? I'll merge it then.

lkoe:
I've made the requested changes.
Please note, however, that this is not a 100% proper fix, as it only there is still a leak (albeit smaller) on project paths and action instances kept in the map.

This fix only helps on keeping the gradle daemon stable for a longer time.

dbmeneses:
Noted @lkoe , I've created a ticket: https://jira.sonarsource.com/browse/SONARGRADL-61.

PR #55, opened 2019-03-28
Clear the actionBroadcastMap once build finishes

lkoe:
Additionally clear the map when the build finishes

### DIFF
--- a/src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
+++ b/src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
@@ -72,6 +72,7 @@ public class SonarQubePlugin implements Plugin<Project> {
       SonarQubeTask sonarQubeTask = project.getTasks().create(SonarQubeExtension.SONARQUBE_TASK_NAME, SonarQubeTask.class);
       sonarQubeTask.setDescription("Analyzes " + project + " and its subprojects with SonarQube.");
       configureTask(sonarQubeTask, project);
+      project.getGradle().buildFinished(p -> actionBroadcastMap.clear());
     }
   }
 


src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -72,6 +72,7 @@ public void apply(Project project) {
       SonarQubeTask sonarQubeTask = project.getTasks().create(SonarQubeExtension.SONARQUBE_TASK_NAME, SonarQubeTask.class);
       sonarQubeTask.setDescription("Analyzes " + project + " and its subprojects with SonarQube.");
       configureTask(sonarQubeTask, project);
+      project.getGradle().buildFinished(p -> actionBroadcastMap.clear());
```
henryju: With this change, is the previous change needed at all?

lkoe: I've not tested it, but it shouldn't be needed strictly. However, I'd consider this change also only to be a workaround.
I am not familiar enough with the inner workings of the plugin to provide a proper fix.

dbmeneses: @lkoe have you tested this change? can 2 builds run in parallel in the same daemon?

lkoe: I've tested it, but not with multiple parallel builds.
As far as I understand the docs (https://docs.gradle.org/current/userguide/gradle_daemon.html) a daemon instance will always only execute one build at a time. If another build is started in parallel gradle starts another daemon instance.

lkoe: In any case, a proper fix for the issue would certainly be preferential.
I've uploaded a snapshot containing the fix above to our local artifactory, so my teams is fine for the time being.

dbmeneses:
thanks again for the contribution

PR #58, opened 2019-04-17
Hadening for Sonar-Gradle

dbmeneses:


dbmeneses:
@henryju The main thing is the memory leak. It's enough to build sonar-enterprise 2-3 times with a gradle daemon and then check a heap dump of the daemon. I can do this if you prefer.

PR #59, opened 2019-04-23
Update jacoco

mezzatto:


henryju:
Hi,

Any reason for this change?

mezzatto:
We have a java 11 project that was unable to use the sonar scanner plugin after being upgraded from java 8. In the latest jacoco release (0.8.3) the official support for java 11 was added (https://github.com/jacoco/jacoco/pull/760).

henryju:
Then you have to update JaCoCo in your project `build.gradle`. The version you changed here is the one used to measure coverage of the scanner source code itself. And since we are still building with Java 8, we don't really need that now.

PR #63, opened 2019-09-27
Fix repositories URLs

jacek-poreda-sonarsource:
Use repox.jfrog

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -51,9 +51,6 @@ ext {
 }
 
 repositories {
-  jcenter()
-  mavenLocal()
-  mavenCentral()
   maven {
     url "https://maven.google.com"
   }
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -45,15 +45,31 @@ if (version.endsWith('-SNAPSHOT') && ext.buildNumber != null) {
...
   jcenter()
   mavenLocal()
   mavenCentral()
   maven {
     url "https://maven.google.com"
   }
+  def repository = project.hasProperty('qa') ? 'sonarsource-qa' : 'sonarsource'
   maven {
-    url "https://repox.sonarsource.com/sonarsource"
+    url "https://repox.jfrog.io/repox/${repository}"
```
simonbrandhof: jcenter and mavenCentral should be removed because they are mirrored by repox.
And are you sure that `mavenLocal` should be kept? It's [not declared in SonarQube](https://github.com/SonarSource/sonarqube/blob/master/build.gradle#L75)

jacek-poreda-sonarsource: I'll try to remove and see if it works

jacek-poreda-sonarsource: @simonbrandhof All right looks like it is working fine without those.

PR #70, opened 2020-03-31
Various tickets

dbmeneses:
GitHub is unstable and because of that Cirrus is having problems so I can't see if the last commit has passing QA, but it was green before that.


henryju:
Don't forget to change version to 3.0-SNAPSHOT

src/main/java/org/sonarqube/gradle/SonarUtils.java
```diff
@@ -41,8 +43,26 @@ static String capitalize(final String word) {
...
+
+    for (Path baseDir : allProjectsBaseDir) {
+      if (!baseDir.toAbsolutePath().normalize().startsWith(rootBaseDir)) {
+        int c1 = baseDir.getNameCount();
+        int c2 = rootBaseDir.getNameCount();
+        Path newBaseDir = rootBaseDir.getRoot();
+        for (int i = 0; i < c1 && i < c2 && baseDir.getName(i).equals(rootBaseDir.getName(i)); i++) {
+          newBaseDir = newBaseDir.resolve(rootBaseDir.getName(i));
+        }
+        rootBaseDir = newBaseDir;
```
henryju, line 62: The downside of this strategy is that it will end up returning `rootBaseDir.getRoot()` if the two paths do not share the same root. Maybe you could ignore all baseDir that are not under the same root than rootBaseDir.getRoot()

src/main/java/org/sonarqube/gradle/SonarUtils.java
```diff
@@ -41,8 +43,26 @@ static String capitalize(final String word) {
...
+    List<Path> allProjectsBaseDir = properties.entrySet().stream()
+      .filter(e -> e.getKey().endsWith(".projectBaseDir"))
+      .map(e -> Paths.get(e.getValue().toString()))
+      .collect(Collectors.toList());
+
+    for (Path baseDir : allProjectsBaseDir) {
+      if (!baseDir.toAbsolutePath().normalize().startsWith(rootBaseDir)) {
+        int c1 = baseDir.getNameCount();
+        int c2 = rootBaseDir.getNameCount();
+        Path newBaseDir = rootBaseDir.getRoot();
```
henryju, line 58:
```suggestion
        if (!rootBaseDir.getRoot().equals(baseDir.getRoot())) {
          LOG.warn("Unable to analyze module " + baseDir + " that is not in the same root");
          continue;
        }
        Path newBaseDir = rootBaseDir.getRoot();
```

src/main/java/org/sonarqube/gradle/SonarUtils.java
henryju: See my suggestion above

PR #72, opened 2020-04-08
Feature/dont depend on test

dbmeneses:
Please review our [contribution guidelines](https://github.com/SonarSource/sonar-scanner-gradle/blob/master/contributing.md).

And please ensure your pull request adheres to the following guidelines: 

- [ ] Please explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- [ ] Use the following formatting style: [SonarSource/sonar-developer-toolset](https://github.com/SonarSource/sonar-developer-toolset#code-style)
- [ ] Provide a unit test for any code you changed
- [ ] If there is a [JIRA](https://jira.sonarsource.com/browse/SONARGRADL) ticket available, please make your commits and pull request start with the ticket ID (SONARGRADL-XXXX)

### DIFF
--- a/integrationTests/src/test/resources/AndroidTestingBlueprint/build.gradle
+++ b/integrationTests/src/test/resources/AndroidTestingBlueprint/build.gradle
@@ -3,14 +3,11 @@
 buildscript {
     repositories {
         jcenter()
-        google()
         mavenLocal()
         maven {
-		      url "https://plugins.gradle.org/m2/"
-		    }
-        maven {
-            url "https://maven.google.com"
+            url "https://plugins.gradle.org/m2/"
         }
+        google()
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:${androidGradle.version}'
@@ -29,13 +26,10 @@ allprojects {
     repositories {
         google()
         jcenter()
-		    maven {
-		        url "https://maven.google.com"
-		    }
     }
 }
 
-taskTree{
+taskTree {
     noRepeat = true  //do not print a sub-tree in the task-tree more than once
     impliesSubProjects = true  //do not print task-tree for child projects in a multi-project
 }
--- a/src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
+++ b/src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
@@ -92,7 +92,7 @@ public class SonarQubePlugin implements Plugin<Project> {
       .filter(p -> p.getPlugins().hasPlugin(JavaPlugin.class) && !p.getExtensions().getByType(SonarQubeExtension.class).isSkipProject())
       .map(p -> p.getTasks().getByName(JavaPlugin.TEST_TASK_NAME))
       .collect(Collectors.toList());
-    sonarQubeTask.dependsOn(testTasks);
+    sonarQubeTask.mustRunAfter(testTasks);
 
     Callable<Iterable<? extends Task>> compileTasks = () -> project.getAllprojects().stream()
       .filter(p -> isAndroidProject(p) && !p.getExtensions().getByType(SonarQubeExtension.class).isSkipProject())
@@ -112,7 +112,7 @@ public class SonarQubePlugin implements Plugin<Project> {
       })
       .flatMap(List::stream)
       .collect(Collectors.toList());
-    sonarQubeTask.dependsOn(compileTasks);
+    sonarQubeTask.mustRunAfter(compileTasks);
   }
 
 }
--- a/src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
+++ b/src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
@@ -21,11 +21,13 @@ package org.sonarqube.gradle;
 
 import com.android.build.gradle.api.BaseVariant;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
@@ -34,6 +36,8 @@ import org.gradle.api.internal.ConventionMapping;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
 import org.gradle.api.plugins.JavaPlugin;
+import org.gradle.testing.jacoco.plugins.JacocoPlugin;
+import org.gradle.testing.jacoco.tasks.JacocoReport;
 
 import static org.sonarqube.gradle.SonarUtils.capitalize;
 import static org.sonarqube.gradle.SonarUtils.isAndroidProject;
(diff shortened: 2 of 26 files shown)


dbmeneses:
I'll create a ticket for the second commit if it's look like a good idea to you, @henryju

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -112,7 +112,7 @@ private void configureTask(SonarQubeTask sonarQubeTask, Project project, Map<Str
       })
       .flatMap(List::stream)
       .collect(Collectors.toList());
-    sonarQubeTask.dependsOn(compileTasks);
+    sonarQubeTask.mustRunAfter(compileTasks);
```
henryju: Even this change could deserve a separate ticket, since it may have impact on accuracy of Java analysis. Bytecode could be missing if people only run `gradle sonarqube`. Are you sure you want to do it?

henryju: My bad, this is only for Android.

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -92,7 +92,7 @@ private void configureTask(SonarQubeTask sonarQubeTask, Project project, Map<Str
       .filter(p -> p.getPlugins().hasPlugin(JavaPlugin.class) && !p.getExtensions().getByType(SonarQubeExtension.class).isSkipProject())
       .map(p -> p.getTasks().getByName(JavaPlugin.TEST_TASK_NAME))
       .collect(Collectors.toList());
-    sonarQubeTask.dependsOn(testTasks);
+    sonarQubeTask.mustRunAfter(testTasks);
```
henryju: I wonder if we should not add a `dependsOn` on the JavaPlugin.COMPILE_JAVA_TASK_NAME/COMPILE_JAVA_TEST_TASK_NAME because contrary to tests, I see no use case where you don't want it have bytecode generated prior to the SQ analysis. It was previously implicit due to the transitive dependency sonarqube -> JavaPlugin.TEST_TASK_NAME -> (COMPILE_JAVA_TASK_NAME, COMPILE_JAVA_TEST_TASK_NAME)

dbmeneses: good point!

integrationTests/src/test/resources/AndroidTestingBlueprint/build.gradle
```diff
@@ -27,11 +26,10 @@ apply plugin: "com.dorongold.task-tree"
...
 allprojects {
     repositories {
-        google()
         jcenter()
-		    maven {
-		        url "https://maven.google.com"
-		    }
+        maven {
+		    url "https://maven.google.com"
+		}
```
vlsi, line 32: The lines have mixed tabs and spaces. Would you please mind to normalize them? (e.g. use spaces as in other lines)

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -88,13 +92,36 @@ private void configureTask(SonarQubeTask sonarQubeTask, Project project, Map<Str
...
+    sonarQubeTask.mustRunAfter(getJacocoTasks(project));
+    sonarQubeTask.dependsOn(getAndroidCompileTasks(project));
+  }
+
+  private static Callable<Iterable<? extends Task>> getJacocoTasks(Project project) {
+    return () -> project.getAllprojects().stream()
+      .filter(p -> p.getPlugins().hasPlugin(JacocoPlugin.class) && !p.getExtensions().getByType(SonarQubeExtension.class).isSkipProject())
+      .map(p -> p.getTasks().withType(JacocoReport.class))
+      .flatMap(Collection::stream)
+      .collect(Collectors.toList());
```
vlsi, line 106: This is not reliable.

1) Tasks might be added even after `sonarqube` plugin is applied. Then the newly added tasks won't be added to the dependency

2) There must not be dependencies among tasks between different projects. See https://docs.gradle.org/current/userguide/cross_project_publications.html#sec:simple-sharing-artifacts-between-projects

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -88,13 +92,36 @@ private void configureTask(SonarQubeTask sonarQubeTask, Project project, Map<Str
...
+  }
+
+  private static Callable<Iterable<? extends Task>> getJavaTestTasks(Project project) {
+    return () -> project.getAllprojects().stream()
       .filter(p -> p.getPlugins().hasPlugin(JavaPlugin.class) && !p.getExtensions().getByType(SonarQubeExtension.class).isSkipProject())
       .map(p -> p.getTasks().getByName(JavaPlugin.TEST_TASK_NAME))
       .collect(Collectors.toList());
```
vlsi, line 113: Same here: please do not query tasks of other projects since it is not reliable (and it might be deprecated completely in the future).

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -88,13 +92,36 @@ private void configureTask(SonarQubeTask sonarQubeTask, Project project, Map<Str
...
-    sonarQubeTask.dependsOn(testTasks);
+  }
 
-    Callable<Iterable<? extends Task>> compileTasks = () -> project.getAllprojects().stream()
+  private static Callable<Iterable<? extends Task>> getJavaCompileTasks(Project project) {
+    return () -> project.getAllprojects().stream()
+      .filter(p -> p.getPlugins().hasPlugin(JavaPlugin.class) && !p.getExtensions().getByType(SonarQubeExtension.class).isSkipProject())
```
vlsi, line 118: The project (and extension) might yet be unconfigured at the time this code is executed.

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
dbmeneses: The `Callable` is resolved only after the project is evaluated, when Gradle needs to figure out what tasks to run and in which order. We've used this for long time without any reported problems. 
About 2., I agree it's not ideal, but because of the way the SonarQube scanner works, it needs to run once for the root project and collect properties for all sub projects. So what we do is apply the plugin to the root project and depend on the subprojects.

vlsi: > after the project is evaluated

The projects might be evaluated in different order.

> it needs to run once for the root project and collect properties for all sub projects. So what we do is apply the plugin to the root project and depend on the subprojects

You might use configurations to share artifacts between projects or implement something like https://github.com/gradle/gradle/tree/master/subprojects/docs/src/samples/jvm-multi-project-with-code-coverage

dbmeneses: Thanks for the example. We can definitely improve things and I will keep it in mind but it's out of scope for these tickets. Here we are just changing the dependencies using the same system as before.

dbmeneses: Also, in my opinion things are not fundamentally different. The main difference is that we try to support multi-project out of the box.

### JaCoCo plugin
The JaCoCo plugin should be applied to every project. This will add report tasks and an extension to the project and to its `test` tasks. The report task sets a `mustRunAfter` the `test`.

The example with the multi-project then recommends that we:
* Create a custom task in the root project
* This task collects data from the JaCoCo extensions added to the `test` task in subprojects
* Add a dependency to the `test` tasks in the subprojects (for which a JaCoCo extension was added)

### Sonarqube plugin
The sonarqube plugin should be applied to a single project. This will add the task to the root project and an extension to all subprojects.

For multi-project, a `dependsOn` or `mustRunAfter` is added to tasks of interest in all subprojects. The sonarqube task will then collect properties from all subprojects by accessing its extension and the other tasks.

PR #83, opened 2021-01-22
SONARGRADL-76 Unwrap Optional

henryju:


src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
```diff
@@ -105,6 +106,7 @@ class FunctionalTests extends Specification {
         def props = new Properties()
         props.load(outFile.newDataInputStream())
         props."sonar.java.jdkHome".contains('8')
+        new File(props."sonar.java.jdkHome").exists()
         props."sonar.java.source" == '8'
         props."sonar.java.target" == '8'
     }
```
m-g-sonar, line 112: I feel that the next test `"set jdkHome, source and target for 'java' projects from task release"` might be missing the same test is missing the test about the actual presence of the property.

Also, what about adding a test targeting misconfiguration? or jdk not found?

src/main/java/org/sonarqube/gradle/SonarUtils.java
```diff
@@ -99,7 +99,7 @@ static void setMainClasspathProps(Map<String, Object> properties, boolean addFor
   }
 
   static void populateJdkProperties(Map<String, Object> properties, JavaCompilerConfiguration config) {
-    properties.put("sonar.java.jdkHome", config.getJdkHome());
+    config.getJdkHome().ifPresent(s -> properties.put("sonar.java.jdkHome", s));
```
m-g-sonar, line 102: Wow, this issue is so bad and unlucky, we should definitely have a rule targeting situation when an `Optional` is provided as argument to a method expecting an `Object` parameter... It is most likely that we would expect to have the content of the `Optional` given, and not an empty `Optional` in worst case scenario.

m-g-sonar (commented):
Overall, it LGTM. The error is unfortunate. I just feel that tests might still be a bit incomplete.

m-g-sonar (approved):
LGTM!

PR #93, opened 2021-05-24
Support Android's Dynamic Features modules

ltpquang:
Current Sonar Scanner Gradle plugin does not scan Android's Dynamic Feature modules. My project got somes dynamic feature modules and they don't show up on my SonarQube dashboard.

This MR intends to add supports for Android Dynamic Feature modules.


henryju (changes requested):
Thanks for the contribution. Would be great to add an integration test if you can, with a sample project having dynamic feature: https://github.com/SonarSource/sonar-scanner-gradle/blob/master/integrationTests/src/test/java/org/sonarqube/gradle/AndroidTest.java

ltpquang:
> Thanks for the contribution. Would be great to add an integration test if you can, with a sample project having dynamic feature: https://github.com/SonarSource/sonar-scanner-gradle/blob/master/integrationTests/src/test/java/org/sonarqube/gradle/AndroidTest.java

@henryju yeah I'm very willing to, but can you show me how I can execute the tests locally, I've played around with `cirrus-qa.sh` but still no progress.

henryju:
@ltpquang Integration tests are in a separate Maven project. You should be able to run them after installing the artifact:
```
./gradlew publishToMavenLocal
cd integrationTests
mvn clean verify
```

ltpquang:
@henryju I managed to get Android tests up and passed, but only some of Gradle tests, still trying to figure them out, can you please take a quick review :D

integrationTests/src/test/resources/android-gradle-dynamic-feature/build.gradle
```diff
@@ -0,0 +1,35 @@
+buildscript {
+    repositories {
+        mavenCentral()
+        mavenLocal()
+        jcenter()
```
henryju, line 5: To be removed

integrationTests/src/test/resources/android-gradle-dynamic-feature/build.gradle
```diff
@@ -0,0 +1,35 @@
...
+}
+
+apply plugin: 'jacoco'
+apply plugin: "org.sonarqube"
+
+allprojects {
+    repositories {
+        mavenCentral()
+        mavenLocal()
+        jcenter()
```
henryju, line 21: To be removed

integrationTests/pom.xml
```diff
@@ -10,14 +10,14 @@
...
-  <version>3.0-SNAPSHOT</version>
+  <version>3.3-SNAPSHOT</version>
   <name>Gradle Plugin :: Integration Tests</name>
   <inceptionYear>2015</inceptionYear>
 
   <properties>
     <!-- Default version when running ITs on dev machines. Value overridden in .cirrus.yml -->
-    <gradle.version>5.6.4</gradle.version>
-    <androidGradle.version>3.3.2</androidGradle.version>
+    <gradle.version>7.0.2</gradle.version>
```
henryju, line 19: I think it might be a bit too much to try to support Gradle 7.0 in our ITs. Many test projects are failing with errors like `Could not find method compile() for arguments [{group=commons-io, name=commons-io, version=2.5}]`
Maybe you should stick to Gradle < 7 for now.

henryju (commented):
@ltpquang This looks really promising. What you have to understand is that we run ITs in a matrix job, against multiple Gradle versions. You can see the combinations here:
https://github.com/SonarSource/sonar-scanner-gradle/blob/eb13f3c668af892a0bb9451ec31f831e4b038fb9/.cirrus.yml#L84

This is very tricky, because from version to version, there are implementation differences, and it may be hard to use the same test project with multiple Gradle/Gradle Android versions. Also the assertions may require to be conditional (the .class output dir changed between Android Gradle versions)

Gradle 7 seems to have a lot of breaking changes, so that is maybe a bit too much to support it right now in our ITs (it might be a separate PR on its own).
At the same time, we are testing very old versions of Gradle, and I think it would be totally fine to stop doing so.

integrationTests/src/test/resources/android-gradle-dynamic-feature/build.gradle
ltpquang: I removed in latest commits

integrationTests/src/test/resources/android-gradle-dynamic-feature/build.gradle
ltpquang: I removed in latest commits

integrationTests/pom.xml
ltpquang: I downgraded to 6.8.3, as it's the latest-below-7 version, will try to resort other test cases

ltpquang: Okay local runs on my machine get all the test passed, please review them one more time @henryju. Thanks so much!

henryju:
The following IT axis are failing:
```
mvn clean verify -Dgradle.version=4.6 -DandroidGradle.version=3.2.1
mvn clean verify -Dgradle.version=5.6.4 -DandroidGradle.version=3.3.2
```
you have to make test assertions conditional to the runtime version of Gradle/Android Gradle

integrationTests/src/test/java/org/sonarqube/gradle/AndroidTest.java
```diff
@@ -50,11 +50,11 @@ public void testAndroidProjectJdk8Retrolambda() throws Exception {
         baseDir.resolve("src/main/res"),
         baseDir.resolve("src/main/AndroidManifest.xml"));
     assertThat(Paths.get(props.getProperty("sonar.tests"))).isEqualTo(baseDir.resolve("src/test/java"));
-    assertThat(Paths.get(props.getProperty("sonar.java.binaries"))).isEqualTo(baseDir.resolve("build/intermediates/javac/debug/compileDebugJavaWithJavac/classes"));
+    assertThat(Paths.get(props.getProperty("sonar.java.binaries"))).isEqualTo(baseDir.resolve("build/intermediates/javac/debug/classes"));
```
henryju, line 53: All those changes are probably related to the update of the Android Gradle plugin to 4.2.0, but since we are running tests in a matrix, you have to use a condition to still support older versions.

ltpquang: ok I got this, let me try a more clean solution to this

ltpquang: maybe JUnit's `assume` mechanism will help us in conditional test running, but fixed versions for conditional switching shouldn't be a clean solution, cannot find out any document or source code that make decision on where to put the intermediates, any suggestion @henryju ?

ltpquang: @henryju  Yes I will update the CI script, the problem is maybe some old Gradle versions are not able to build dynamic features module, in such a case I intend to:

- Revert `AndroidTestingBlueprint` to original state
- Clone another blue print project which has dynamic feature

After that, the old test cases will skip on new Gradle version, and the new test cases will run only new Gradle version. How do you think?

henryju: That sound very good to me :)

ltpquang: @henryju I updated test cases with some conditional checks as discussed, could you please take another review?

henryju:
Hi @ltpquang 

The changes look good. I have created an internal PR to trigger the entire pipeline, and there are some errors in the IT axis for Gradle 4.6 and Android Gradle 3.2.1:
```
org.gradle.api.ProjectConfigurationException: A problem occurred configuring project ':mydynamicfeature'.
[...]
Caused by: com.android.builder.errors.EvalIssueException: Default file proguard-android-optimize.txt should not be specified in this module. It can be specified in the base module instead.
	at com.android.build.gradle.internal.tasks.MergeConsumerProguardFilesConfigAction.checkForDefaultFiles(MergeConsumerProguardFilesConfigAction.java:130)
[...]
BUILD FAILED in 16s
Tests run: 10, Failures: 0, Errors: 1, Skipped: 2, Time elapsed: 138.393 sec <<< FAILURE! - in org.sonarqube.gradle.AndroidTest
testAndroidDynamicFeature(org.sonarqube.gradle.AndroidTest)  Time elapsed: 16.837 sec  <<< ERROR!
[...]
Tests in error: 
  AndroidTest.testAndroidDynamicFeature:125->AbstractGradleIT.runGradlewSonarQubeSimulationMode:64->AbstractGradleIT.runGradlewSonarQubeSimulationModeWithEnv:68->AbstractGradleIT.runGradlewSonarQubeSimulationModeWithEnv:77->AbstractGradleIT.runGradlewSonarQubeWithEnv:94 » Runtime
Tests run: 23, Failures: 0, Errors: 1, Skipped: 2
```
https://cirrus-ci.com/task/5946464187711488?logs=main

ltpquang:
@henryju thanks Henry, do you have any estimation on when will the next release come out?

henryju:
I just released 3.3 :)

mochadwi:
Is there any example to implement sonarqube for multi-module with Android Dynamic Feature (DFM) + Kotlin/Java Library?

```
|-domain // kotlin
|-data // kotlin library
|-app // android library
|-dfm // contains sub project of android dynamic feature library
|-----/dfm-a
|-----/dfm-b
```

ltpquang:
> Is there any example to implement sonarqube for multi-module with Android Dynamic Feature (DFM) + Kotlin/Java Library?
> 
> ```
> |-domain // kotlin
> |-data // kotlin library
> |-app // android library
> |-dfm // contains sub project of android dynamic feature library
> |-----/dfm-a
> |-----/dfm-b
> ```

@mochadwi Please open a separated issue/ SO question, since it's not related to the PR.

mochadwi:
for future reader seems related: https://stackoverflow.com/a/60036279/3763032

PR #100, opened 2021-11-19
Replace deprecated Report method usage

G00fY2:
This PR may serve as a guidance on how to fix the following issue: https://community.sonarsource.com/t/report-enabled-and-report-destination-are-deprecated-and-will-be-removed-in-gradle-8-0/50491/3

`getRequired()` is marked as _Incubating_ in Gradle 6.7. But I guess you're already planing on migrating to Gradle 7.x.

### DIFF
--- a/src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
+++ b/src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
@@ -248,8 +248,8 @@ public class SonarPropertyComputer {
   private static void configureJaCoCoCoverageReport(final Test testTask, final boolean addForGroovy, Project project, final Map<String, Object> properties) {
     project.getTasks().withType(JacocoReport.class, jacocoReportTask -> {
       SingleFileReport xmlReport = jacocoReportTask.getReports().getXml();
-      if (xmlReport.isEnabled() && xmlReport.getDestination().exists()) {
-        appendProp(properties, "sonar.coverage.jacoco.xmlReportPaths", xmlReport.getDestination());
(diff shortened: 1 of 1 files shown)


wouter-admiraal-sonarsource:
Hi @G00fY2,

Thank you for your contribution.

As of now, we're not planning any immediate updates on our Gradle scanner, so no, right now there's no plan to migrate to Gradle 7.x. That being said, we will have a look at your fix 🙂.

wouter-admiraal-sonarsource:
We had a quick look and from what we understand, this change will break support of older Gradle versions. Right now, this isn't something we take lightly. Probably we would have to use some reflection or runtime version check to use the right method, rather than simply drop support for Gradle 6.

I've opened a ticket to track this: https://jira.sonarsource.com/browse/SONARGRADL-84

For the time being, I will close this PR. Thanks again for taking the time to suggest his change, it's much appreciated 🙂.

agabrys:
> rather than simply drop support for Gradle 6.

You don't have to stop supporting Gradle 6.X. The replacement methods have been introduced in Gradle 6.1:
* `getDestination()` → `getOutputLocation()`
* `isEnabled()` → `getRequired().get()`
* `setEnabled​(boolean enabled)` → `getRequired().set(boolean enabled)`

See the [org.gradle.api.reporting.Report 7.3.3](https://docs.gradle.org/7.3.3/javadoc/org/gradle/api/reporting/Report.html) documentation.

Gradle 6.1 has been released at Jan 15, 2020 (almost 2 years old). There were only two 6.X versions released earlier:
* 6.0 - Nov 08, 2019
* 6.0.1 - Nov 18, 2019

I think developers shouldn't have a problem with upgrading from 6.0 or 6.0.1 to 6.1 🙂

PR #106, opened 2022-03-24
SONARGRADL-89 Add java 17 in CI to sonar-scanner-gradle

zipeng-wu-sonarsource:


build.gradle
```diff
@@ -69,13 +69,13 @@ repositories {
 }
 
 dependencies {
-  implementation 'org.sonarsource.scanner.api:sonar-scanner-api:2.16.1.361'
+  implementation 'org.sonarsource.scanner.api:sonar-scanner-api:2.16.2.588'
```
dbmeneses, line 72: this is potentially a functional change, doesn't affect tests only.
Was it necessary for it to run with java 17?
If not, it should be done with a dedicated commit/ticket.

integrationTests/src/test/java/org/sonarqube/gradle/AbstractGradleIT.java
```diff
@@ -129,7 +129,7 @@ protected RunResult runGradlewWithEnvQuietly(String project, String exeRelativeP
       .directory(exeDir)
       .redirectOutput(outputFile)
       .redirectErrorStream(true);
-    pb.environment().put("GRADLE_OPTS", "-Xmx1024m");
+    pb.environment().put("GRADLE_OPTS", "-Xmx1024m --add-opens=java.prefs/java.util.prefs=ALL-UNNAMED");
```
dbmeneses, line 132: Does it mean that to run the scanner with java 17 we need to add this JVM parameter?
What fails if this is not passed?

zipeng-wu-sonarsource: It had this issue with Java 17(and gradle 7.3+ doesn't fix it): https://github.com/gradle/gradle/issues/18508

build.gradle
zipeng-wu-sonarsource: yeah indeed, this update might not be neccessary, i will roll it back

dbmeneses:
@zipeng-wu-sonarsource The commit ticket number/message should be updated to https://jira.sonarsource.com/browse/SONARGRADL-89

dbmeneses (approved):
great job 👍

PR #109, opened 2022-03-29
Do not use deprecated Report methods backwards compatibly

henryju:
Same as https://github.com/SonarSource/sonar-scanner-gradle/pull/108 but with QA


henryju:
@dbmeneses May I ask you to double check the changes. The initial commit comes from https://github.com/SonarSource/sonar-scanner-gradle/pull/108 but I have also improved/cleaned the QA.

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
```diff
@@ -270,7 +273,7 @@ private static void configureJaCoCoCoverageReport(final Test testTask, final boo
   }
 
   private static void configureTestReports(Test testTask, Map<String, Object> properties) {
-    File testResultsDir = testTask.getReports().getJunitXml().getDestination();
+    File testResultsDir = getDestination(testTask.getReports().getJunitXml());
```
dbmeneses, line 277: `getDestination` should be annotated with `CheckForNull` and the following if should check for null.

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
```diff
@@ -248,8 +250,9 @@ private static void extractTestProperties(Project project, Map<String, Object> p
   private static void configureJaCoCoCoverageReport(final Test testTask, final boolean addForGroovy, Project project, final Map<String, Object> properties) {
     project.getTasks().withType(JacocoReport.class, jacocoReportTask -> {
       SingleFileReport xmlReport = jacocoReportTask.getReports().getXml();
-      if (xmlReport.isEnabled() && xmlReport.getDestination().exists()) {
-        appendProp(properties, "sonar.coverage.jacoco.xmlReportPaths", xmlReport.getDestination());
+      File reportDestination = getDestination(xmlReport);
+      if (isReportEnabled(xmlReport) && reportDestination.exists()) {
```
dbmeneses: check for null here too

dbmeneses (approved):
Great clean up!
Would be good to add the ticket number to the first commit's message

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
henryju: Already done in last commit, or am I missing something?

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
henryju: Already done in last commit, or am I missing something?

PR #111, opened 2022-04-15
BUILD-1449 sbom plugin

julien-carsique-sonarsource:
BUILD-1449: use org.cyclonedx.bom Gradle plugin to generate a SBOM artifact attached to the build


tomverin (approved):

julien-carsique-sonarsource:
PR updated to include only the runtime dependencies, ie: sonar-scanner-api

henryju (approved):

PR #123, opened 2022-10-17
[NO JIRA] Fix README regarding using composite builds

dbmeneses:
The other way also works (it's the old way of applying plugins), but it required also to use a `buildscript` block to include the plugin in the classpath.


lukasz-jarocki-sonarsource (approved):
Including buildscript indeed seems to help

PR #126, opened 2022-10-18
SONARGRADL-93 Rename the task "sonarqube" to "sonar"

dbmeneses:


src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -45,17 +45,15 @@
 import static org.sonarqube.gradle.SonarUtils.isAndroidProject;
 
 /**
- * A plugin for analyzing projects with the <a href="http://redirect.sonarsource.com/doc/analyzing-with-sq-gradle.html">SonarQube Scanner</a>.
+ * A plugin for analyzing projects with the <a href="http://redirect.sonarsource.com/doc/analyzing-with-sq-gradle.html">SonarScanner for Gradle</a>.
  * When applied to a project, both the project itself and its subprojects will be analyzed (in a single run).
  */
 public class SonarQubePlugin implements Plugin<Project> {
```
leo-geoffroy-sonarsource, line 51: Can we rename the classname too?

leo-geoffroy-sonarsource:
There might some changes in the general build.gradle:
build.gradle:21 -> 'SonarQube Scanner for Gradle'
build.gradle:85
`gradlePlugin {
  plugins {
    sonarqubePlugin {
      displayName = projectName
      description = project.description
      id = 'org.sonarqube'
      implementationClass = 'org.sonarqube.gradle.SonarQubePlugin'
    }
  }
}`

That raises the question if we want to publish the plugin  under a different id ?

dbmeneses:
> There might some changes in the general build.gradle: build.gradle:21 -> 'SonarQube Scanner for Gradle' build.gradle:85
> 
> ```
>   plugins {
>     sonarqubePlugin {
>       displayName = projectName
>       description = project.description
>       id = 'org.sonarqube'
>       implementationClass = 'org.sonarqube.gradle.SonarQubePlugin'
>     }
>   }
> }
> ```
> 
> That raises the question if we want to publish the plugin under a different id ?

That's the question I raised in the ticket. I think that would have a high impact since we can't just deprecated the old id. I'm also not sure how it would affect our plugin in the gradle plugin repository. I suggest we leave that as is.

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
dbmeneses: I think it only makes sense to rename if we actually rename the plugin id

leo-geoffroy-sonarsource:
> > There might some changes in the general build.gradle: build.gradle:21 -> 'SonarQube Scanner for Gradle' build.gradle:85
> > ```
> >   plugins {
> >     sonarqubePlugin {
> >       displayName = projectName
> >       description = project.description
> >       id = 'org.sonarqube'
> >       implementationClass = 'org.sonarqube.gradle.SonarQubePlugin'
> >     }
> >   }
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That raises the question if we want to publish the plugin under a different id ?
> 
> That's the question I raised in the ticket. I think that would have a high impact since we can't just deprecated the old id. I'm also not sure how it would affect our plugin in the gradle plugin repository. I suggest we leave that as is.

Ok I see the point, It would mean publishing twice the artifacts, that would be a lot more complex. I agree on just proceeding with task/extension renaming for now.

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
leo-geoffroy-sonarsource: make sense

leo-geoffroy-sonarsource (approved):

PR #128, opened 2022-10-19
SONARGRADL-95 added android properties to scanner when scanning andro…

lukasz-jarocki-sonarsource:
SONARGRADL-95

### DIFF
--- a/src/main/java/org/sonarqube/gradle/AndroidUtils.java
+++ b/src/main/java/org/sonarqube/gradle/AndroidUtils.java
@@ -21,6 +21,7 @@ package org.sonarqube.gradle;
 
 import com.android.build.gradle.AppExtension;
 import com.android.build.gradle.AppPlugin;
+import com.android.build.gradle.BaseExtension;
 import com.android.build.gradle.DynamicFeaturePlugin;
 import com.android.build.gradle.LibraryExtension;
 import com.android.build.gradle.LibraryPlugin;
@@ -33,6 +34,8 @@ import com.android.build.gradle.api.UnitTestVariant;
 import com.android.build.gradle.internal.api.TestedVariant;
 import com.android.build.gradle.internal.tasks.DeviceProviderInstrumentTestTask;
 import com.android.build.gradle.tasks.factory.AndroidUnitTest;
+import com.android.build.gradle.internal.dsl.ProductFlavor;
+import com.android.builder.model.ApiVersion;
 import com.android.builder.model.SourceProvider;
 import java.io.File;
 import java.lang.reflect.InvocationTargetException;
@@ -44,9 +47,11 @@ import java.util.LinkedHashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.annotation.Nullable;
 import org.gradle.api.Project;
 import org.gradle.api.file.DirectoryProperty;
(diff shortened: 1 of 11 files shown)


src/main/java/org/sonarqube/gradle/AndroidUtils.java
```diff
@@ -147,27 +178,32 @@ private static String getTestBuildType(Project project) {
...
   @Nullable
-  static BaseVariant findVariant(Project project, @Nullable String userConfiguredBuildVariantName) {
+  static AndroidVariantAndExtension findVariantAndExtension(Project project, @Nullable String userConfiguredBuildVariantName) {
     String testBuildType = getTestBuildType(project);
     PluginCollection<AppPlugin> appPlugins = project.getPlugins().withType(AppPlugin.class);
     if (!appPlugins.isEmpty()) {
       AppExtension androidExtension = project.getExtensions().getByType(AppExtension.class);
-      return findVariant(new ArrayList<>(androidExtension.getApplicationVariants()), testBuildType, userConfiguredBuildVariantName);
+      BaseVariant variant = findVariant(new ArrayList<>(androidExtension.getApplicationVariants()), testBuildType, userConfiguredBuildVariantName);
+      return new AndroidVariantAndExtension(androidExtension, variant, userConfiguredBuildVariantName);
```
leo-geoffroy-sonarsource, line 187: userConfiguredBuildVariantName is not customized in this method, could we remove it from AndroidVariantAndExtension

src/main/java/org/sonarqube/gradle/AndroidUtils.java
```diff
@@ -101,6 +108,30 @@ private static void configureForAndroid(Project project, BaseVariant variant, Ma
     }
   }
 
+  private static void populateSonarQubeAndroidProperties(AndroidVariantAndExtension android, Map<String, Object> properties) {
```
leo-geoffroy-sonarsource, line 111: I propose populateSonarQubeSdkVersionAndroidProperties  since we are populating android props elsewhere too

src/main/java/org/sonarqube/gradle/AndroidUtils.java
```diff
@@ -101,6 +108,30 @@ private static void configureForAndroid(Project project, BaseVariant variant, Ma
...
+    if (android.isVariantProvidedByUser()) {
+      ApiVersion minSdkVersion = android.getVariant().getMergedFlavor().getMinSdkVersion();
+      if (minSdkVersion != null) {
+        properties.put(AndroidProperties.MIN_SDK_VERSION_MIN, minSdkVersion.getApiLevel());
+        properties.put(AndroidProperties.MIN_SDK_VERSION_MAX, minSdkVersion.getApiLevel());
+      }
+    } else {
+      Set<Integer> minSdks = android.getExtension().getProductFlavors().stream().map(ProductFlavor::getMinSdk).collect(Collectors.toSet());
+      minSdks.add(android.getExtension().getDefaultConfig().getMinSdk());
+      minSdks.remove(null);
```
leo-geoffroy-sonarsource, line 127: Can we instead don't add the null entry instead of deleting it afterward

lukasz-jarocki-sonarsource: We could. This would require us to check both in line 125 and in line 126 for null or write our own implementation of Set that ignores nulls.

src/main/java/org/sonarqube/gradle/AndroidUtils.java
lukasz-jarocki-sonarsource: Not sure I understand 🧐 We need it for this logic:
      this.variantProvidedByUser = variant != null && variant.getName().equals(userConfiguredVariantName);
that resides in the constructor of AndroidVariantAndExtension.

leo-geoffroy-sonarsource: It is passed in the method @Nullable String userConfiguredBuildVariantName, but returned in the output object.
But with a second look, it seems fine with usage of  the method isVariantProvidedByUser

src/main/java/org/sonarqube/gradle/AndroidUtils.java
leo-geoffroy-sonarsource: I think that immutability is preferred when we can
Other solution 
```java
Stream.concat ( android.getExtension().getProductFlavors().stream(), Stream.of (ndroid.getExtension().getDefaultConfig().getMinSdk()))
.filter(Objects::nonNull)
.collect(Collectors.toSet())
```

I recon this does not make it necessarily clearer

lukasz-jarocki-sonarsource: Fixed

leo-geoffroy-sonarsource (approved):

PR #129, opened 2022-10-19
SONARGRADL-90 Use the ProviderAPI to allow lazy configuration

dbmeneses:
Please ignore the first commit, which is part of #126 . I'd have conflicts if I didn't include it.


src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -96,10 +94,8 @@ private static void addExtensions(Project project, String name, Map<String, Acti
   }
 
   private static void configureTask(SonarTask sonarTask, Project project, Map<String, ActionBroadcast<SonarProperties>> actionBroadcastMap) {
-    ConventionMapping conventionMapping = sonarTask.getConventionMapping();
-    // this will call the SonarPropertyComputer to populate the properties of the task just before running it
-    conventionMapping.map("properties", () -> new SonarPropertyComputer(actionBroadcastMap, project)
-      .computeSonarProperties());
+    sonarTask.properties = project.provider(() -> new SonarPropertyComputer(actionBroadcastMap, project).computeSonarProperties())
```
leo-geoffroy-sonarsource, line 97: maybe a setter for "sonarTask.properties" and field in private?

leo-geoffroy-sonarsource (approved):

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
dbmeneses: ok, fixed.

PR #135, opened 2022-10-24
SONARGRADL-26 Automatically configure sonar.junit.reportsPath on Andr…

dbmeneses:


lukasz-jarocki-sonarsource:
Don't we need to write unit tests for this logic?

dbmeneses:
> Don't we need to write unit tests for this logic?

We have ITs for it. Unit tests would pretty much replicate what we have in the ITs, unless we mock all the objects involved, which I think would be too much work.
Unfortunately I don't think we can run instrumentation tests in the ITs.

src/main/java/org/sonarqube/gradle/AndroidUtils.java
```diff
@@ -101,6 +108,49 @@ private static void configureForAndroid(Project project, BaseVariant variant, Ma
...
+    if (directories.isEmpty()) {
+      return;
+    }
+
+    String value = directories.stream()
+      .filter(Provider::isPresent)
+      .map(d -> d.get().getAsFile())
+      .filter(file -> file.isDirectory() && file.list().length > 0)
+      .map(File::getAbsolutePath)
+      .collect(Collectors.joining(","));
```
leo-geoffroy-sonarsource, line 150: We could put a list of files instead, it is resolve later by : org/sonarqube/gradle/SonarPropertyComputer.java:173

src/main/java/org/sonarqube/gradle/AndroidUtils.java
```diff
@@ -101,6 +108,49 @@ private static void configureForAndroid(Project project, BaseVariant variant, Ma
...
+
+    if (directories.isEmpty()) {
+      return;
+    }
+
+    String value = directories.stream()
+      .filter(Provider::isPresent)
+      .map(d -> d.get().getAsFile())
+      .filter(file -> file.isDirectory() && file.list().length > 0)
+      .map(File::getAbsolutePath)
```
leo-geoffroy-sonarsource, line 149: Check if file exists?

leo-geoffroy-sonarsource, line 149: Check if file exists?

leo-geoffroy-sonarsource:
Don't know if it relates too much, but Variants (incl TestVariant) are [deprecated ](https://developer.android.com/studio/releases/gradle-plugin-roadmap#agp_90_first_half_of_2023) and will be removed in Android gradle plugin 9
Also [reference documentation](https://medium.com/androiddevelopers/new-apis-in-the-android-gradle-plugin-f5325742e614#e7e5)

This means we will need to support the new variant mechanism also..

src/main/java/org/sonarqube/gradle/AndroidUtils.java
```diff
@@ -101,6 +108,49 @@ private static void configureForAndroid(Project project, BaseVariant variant, Ma
     }
   }
 
+  private static void configureTestReports(Project project, BaseVariant variant, Map<String, Object> map) {
+    if (!(variant instanceof TestedVariant)) {
+      return;
+    }
+
+    if (GradleVersion.current().compareTo(GradleVersion.version("6.0")) < 0) {
```
leo-geoffroy-sonarsource, line 116: Should it depend on android gradle plugin version instead (com.android.tools.build:gradle), since TestedVariant is part of this package?

src/main/java/org/sonarqube/gradle/AndroidUtils.java
dbmeneses: it's checking that it's a directory and that it's not empty

src/main/java/org/sonarqube/gradle/AndroidUtils.java
dbmeneses: ok, I'll fix

dbmeneses: actually since it's multiple files, I don't think that would work

dbmeneses: but I'll use `getEscapedFilePath`

leo-geoffroy-sonarsource: Multiple file is fine normally since org.sonarqube.gradle.SonarPropertyComputer#convertValue support List<File>

leo-geoffroy-sonarsource (commented):
LGTM, just need to adapt the version check with the new class and maybe use convertValue as it supports Iterables

src/main/java/org/sonarqube/gradle/AndroidUtils.java
dbmeneses: ok, I'll try

leo-geoffroy-sonarsource (approved):

PR #137, opened 2022-10-25
SONARGRADL-90 Use the ProviderAPI to allow lazy configuration

dbmeneses:
Avoid multiple computations of the properties, which may result in repeated logging.


dbmeneses:
I think ITs are failing because of #136. Please rerun ITs once that is merged.

leo-geoffroy-sonarsource:
It doesn't run twice anymore, but it seems like it is incompatible with the oldest version of gradle

leo-geoffroy-sonarsource (approved):

PR #138, opened 2022-10-26
SONARGRADL-24 Add automatic detection android lint

leo-geoffroy-sonarsource:
I have added a new test as I needed to run the lint command on the projects and this requires  up-to-date structure of android project with higher minSdkVersions.

I also needed to have more than one subprojects to test different scenarios


src/main/java/org/sonarqube/gradle/AndroidUtils.java
```diff
@@ -65,7 +66,10 @@
  */
 class AndroidUtils {
 
+  private static final String LINT_REPORT_STANDARD_PATH = "reports/lint-results-%s.xml";
+  private static final String LINT_REPORT_DEFAULT_PATH = "reports/lint-results.xml";
```
dbmeneses, line 70: No longer used?

src/main/java/org/sonarqube/gradle/AndroidUtils.java
```diff
@@ -74,11 +78,33 @@ static void configureForAndroid(Project project, String userConfiguredBuildVaria
     BaseVariant variant = findVariant(project, userConfiguredBuildVariantName);
     if (variant != null) {
       configureForAndroid(project, variant, properties);
+      configureAndroidLintReport(project, variant, properties);
```
dbmeneses, line 81: Wouldn't it be better placed in `configureForAndroid` since it's part of the android configuration?

dbmeneses (approved):

src/main/java/org/sonarqube/gradle/AndroidUtils.java
leo-geoffroy-sonarsource: yes make sense, especially with the addition of configureTestReports

PR #139, opened 2022-10-26
[NO JIRA] parse android plugin version

dbmeneses:


src/main/java/org/sonarqube/gradle/SonarUtils.java
```diff
@@ -34,6 +34,8 @@
 import javax.annotation.Nullable;
 import org.gradle.api.Project;
 
+import static com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION;
```
lukasz-jarocki-sonarsource, line 37: This is not the right constant. I will fix it when I come to the office.

lukasz-jarocki-sonarsource: Pushed the right, deprecated constant. Although I honestly think we shouldn't import Android related classes in SonarUtils.

lukasz-jarocki-sonarsource (approved):

PR #146, opened 2023-01-22
Compatibility with upcoming gradle 8.0.

cocorossello:
This PR provides compatibility with upcoming gradle 8.0.

Note that the required minimum version must be bumped to 6.1 with this patch (report.isEnabled is removed in gradle 8.0, some reflection magic would be required to support gradle < 6.1)

Tested with gradle 6.1, 7.6 and 8.0-RC2

### DIFF
--- a/src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
+++ b/src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
@@ -48,6 +48,7 @@ import org.gradle.api.plugins.JavaPlugin;
 import org.gradle.api.plugins.JavaPluginConvention;
 import org.gradle.api.reporting.Report;
 import org.gradle.api.reporting.SingleFileReport;
+import org.gradle.api.reporting.internal.SimpleReport;
 import org.gradle.api.tasks.SourceSet;
 import org.gradle.api.tasks.SourceSetOutput;
 import org.gradle.api.tasks.compile.JavaCompile;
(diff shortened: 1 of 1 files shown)


src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
```diff
@@ -410,23 +411,20 @@ private String computeProjectKey() {
   }
 
   private static boolean isReportEnabled(Report report) {
-    if (GradleVersion.version("7.0").compareTo(GradleVersion.current()) <= 0) {
-      return report.getRequired().getOrElse(false);
-    } else {
-      return report.isEnabled();
-    }
+    return report.getRequired().getOrElse(false);
```
lukasz-jarocki-sonarsource, line 414: Do you have a good understanding how is this approach working both with Gradle 7 and Gradle 8 at the same time?
The return type is different on Gradle 7 and 8 for the method `report.getRequired()`. So if this code gets compiled with Gradle 8 it should not work anymore if the user chooses to use Gradle 7.

lukasz-jarocki-sonarsource:
Hi @cocorossello ,
We are considering this PR as well as discussing whether we need to support Gradle 5.
Casting to SimpleReport is a little bit of a hack and we are looking into other solutions too (Reflection).

We will update the status of this PR as soon as we finish discussions.

cocorossello:
I can see that you solved it covering more cases.

I provided this PR because I thought it was ok to remove support for gradle 6.1, your solution is more compatible. Closing

PR #147, opened 2023-01-24
SONARGRADL-103 Update license headers for 2023

wouter-admiraal-sonarsource:


stanislavhh:
I think we also need to update `./NOTICE.txt` file (and add this step to the doc)

stanislavhh (approved):

PR #149, opened 2023-02-03
Support Gradle 8

lukasz-jarocki-sonarsource:
1. I've tried introducing some test for SonarPropertyComputer (doesn't have any test yet) and I failed. Since we are relying heavily on reflection and Gradle version - it is not so easy to test.
2. Android Gradle Plugin 7.x is not compatible with Gradle 8. Android Gradle Plugin 8.x is not released yet. 
https://mvnrepository.com/artifact/com.android.tools.build/gradle?repo=google

So if we merge this PR in its current form we would need to add a ticket to bump AGP when they release it (and add IT for it).

### DIFF
--- a/src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
+++ b/src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
@@ -22,6 +22,8 @@ package org.sonarqube.gradle;
 import java.io.File;
 import java.io.IOException;
 import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import org.gradle.api.provider.Provider;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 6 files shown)


rpalcolea:
this is amazing! would be great if it could be released soon! :D

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
```diff
@@ -410,23 +412,54 @@ private String computeProjectKey() {
...
+  /**
+   * For Gradle 7+
+   */
+  private static File getDestinationNewApi(Report report) throws NoSuchMethodException, InvocationTargetException, IllegalAccessException {
+    Provider<? extends FileSystemLocation> provider;
+    if (GradleVersion.version("8.0").compareTo(GradleVersion.current()) <= 0) {
+      provider = report.getOutputLocation();
     } else {
-      return report.getDestination();
+      Method getOutputLocationGradle7 = report.getClass().getMethod("getOutputLocation");
```
dbmeneses, line 448: I think I'm missing something here - this looks like the same method called for v8.0?

dbmeneses (approved):

dbmeneses:
Could we run the ITs with Gradle 8, disabling the android tests?

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
lukasz-jarocki-sonarsource: Almost the same. The return type has changed. So for Java these are two different methods of course.

More details: https://github.com/gradle/gradle/pull/22232/files

lukasz-jarocki-sonarsource:
> Could we run the ITs with Gradle 8, disabling the android tests?

Ok, I will do that tomorrow.

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
dbmeneses: ok, got it

lukasz-jarocki-sonarsource:
Ticket created for AGP: https://sonarsource.atlassian.net/browse/SONARGRADL-107

dbmeneses:
@rpalcolea we might wait for gradle 8's final release.

rpalcolea:
And 8.0 is out 😎 would be great to have a release with this soon.

thanks again for the work to support new gradle versions so quickly!

PR #150, opened 2023-02-06
SONARGRADL-100 sonar task fails when debug logs are enabled

dbmeneses:


integrationTests/src/test/java/org/sonarqube/gradle/GradleTest.java
```diff
@@ -53,6 +54,14 @@ public void testSimpleJavaProject() throws Exception {
     assertThat(props.getProperty("sonar.java.test.libraries")).contains(baseDir.resolve("build/classes/java/main").toString());
   }
 
+  /**
+   * SONARGRADL-100
+   */
+  @Test
+  public void testDebugModeEnabled() throws Exception {
+    runGradlewSonarSimulationModeWithEnv("/java-gradle-simple", emptyMap(), "-d");
```
tom-vanbraband-sonarsource, line 62: To make the intent of the test clearer here you can wrap it with assertj `assertThatCode.doesNotThrowAnyException()`:
```suggestion
assertThatCode(() ->  runGradlewSonarSimulationModeWithEnv("/java-gradle-simple", emptyMap(), "-d"))
    .doesNotThrowAnyException();
```

lukasz-jarocki-sonarsource: I honestly don't feel we need to do that.

We could just write @Test(expected = Test.None.class) which is built-in method in JUnit 4 to express what kind of exception is expected (or no exception). Even better, the Test.None.class is default value for `expected` so we can omit that and we end up with just `@Test`

In the same way as when declaring a variable in Java we don't write 
`private File file = null;`
but instead we write
`private File file;`
as the null value is the default one.

lukasz-jarocki-sonarsource (approved):

integrationTests/src/test/java/org/sonarqube/gradle/GradleTest.java
dbmeneses: I see both points.. I agree that we don't **need** to do it but I like the idea of making the intention clear in this case. I'll add it.

PR #156, opened 2023-02-15
NO-JIRA moving from release candidate to released version of gradle

lukasz-jarocki-sonarsource:
The failure of the QA in unexpected. I will look into that.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -144,7 +144,7 @@ gradle_qa_java17_task:
     matrix:
       - GRADLE_VERSION: 7.4.1
         ANDROID_GRADLE_VERSION: 7.1.0
-      - GRADLE_VERSION: 8.0-rc-3
+      - GRADLE_VERSION: 8.0
   gradle_cache:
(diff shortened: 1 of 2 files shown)


dbmeneses:
@lukasz-jarocki-sonarsource This is what I warned about before - the unit tests are not compatible with gradle 8. It used to work because cirrus used gradle (and not gradlew) to build and run the tests, so the version was hardcoded to 7.4 (which is what is shipped in the docker image being used). Now it uses gradlew and unit tests fail. It's just a few things in the code of the unit tests themselves.

lukasz-jarocki-sonarsource:
> @lukasz-jarocki-sonarsource This is what I warned about before - the unit tests are not compatible with gradle 8. It used to work because cirrus used gradle (and not gradlew) to build and run the tests, so the version was hardcoded to 7.4 (which is what is shipped in the docker image being used). Now it uses gradlew and unit tests fail. It's just a few things in the code of the unit tests themselves.

Makes sense.

PR #159, opened 2023-02-16
NO-JIRA moving from release candidate to released version of gradle

lukasz-jarocki-sonarsource:


lukasz-jarocki-sonarsource:
Seems like we can't build on Cirrus with Gradle 8. This is because one of the gradle tasks for this build is running sonarqube which of course runs sonar-scanner-gradle. But it is built version of sonar-scanner-gradle - at this moment is 3.5.0.2730 which is not compatible with Gradle 8 (new method that we use (getOutputLocation()) when running with Gradle 8 was not called of course in 3.5.0.2730 as it didn't exist yet).

So my suggestion is to stick to build with Gradle 7 for now until Scanner for Gradle 4 is released.

I made the unit tests compatible with both Gradle 7 and 8.

dbmeneses:
sounds good to me. Since we have ITs against Gradle 8 I see no strong need to move the build to Gradle 8.

dbmeneses (approved):

lukasz-jarocki-sonarsource:
Tested manually on 2 different repos with versions of Gradle 5, 7 and 8 and works without any problems.

PR #155, opened 2023-02-14
Configure tasks on demand

Sineaggi:
As per https://melix.github.io/blog/2022/05/gradle-laziness.html, we'll try to avoid eagerly configuring the task unless the user needs it.
Currently the sonarqube gradle plugin does not support gradle 8. Instead of disabling the relevant tasks, we have to remove the plugin as the tasks are eagerly configured.
```
java.lang.NoSuchMethodError: 'org.gradle.api.provider.Provider org.gradle.api.reporting.Report.getOutputLocation()'
	at org.sonarqube.gradle.SonarPropertyComputer.getDestination(SonarPropertyComputer.java:423)
	at org.sonarqube.gradle.SonarPropertyComputer.configureTestReports(SonarPropertyComputer.java:286)
	at org.sonarqube.gradle.SonarPropertyComputer.extractTestProperties(SonarPropertyComputer.java:255)
	at org.sonarqube.gradle.SonarPropertyComputer.lambda$configureForJava$5(SonarPropertyComputer.java:226)
	at org.gradle.api.internal.collections.CollectionFilter$1.execute(CollectionFilter.java:59)
```

PR #161, opened 2023-03-06
Configure tasks on demand

lukasz-jarocki-sonarsource:
External PR: https://github.com/SonarSource/sonar-scanner-gradle/pull/155

Tested on Gradle 5 and it works. More details: https://melix.github.io/blog/2022/05/gradle-laziness.html


dbmeneses:
The `SonarPropertyComputer` is called lazily when the `sonar` test executes. At that time, we want to make sure that the property map is populated with data from all tasks for all subprojects, even if it means to eagerly initialize some of those tasks. Note that we are not lazily configuring those tasks, we are extracting information from them/
So I've some doubts about this PR.

lukasz-jarocki-sonarsource:
Closing this as:
- there is no need for that as of now
- we don't have enough evidence that in our case it brings any performance improvement

PR #163, opened 2023-05-15
SONARGRADL-110 Set property `sonar.java.enablePreview` when Java's enablePreview feature is set

kaufco:


src/main/java/org/sonarqube/gradle/JavaCompilerUtils.java
```diff
@@ -89,6 +89,7 @@ private static boolean areToolchainsSupported() {
   // Inspired by
   // https://github.com/gradle/gradle/blob/d3e4faca3df507176b67d9b3bb3ee91bf2aa070c/subprojects/language-java/src/main/java/org/gradle/api/tasks/compile/JavaCompile.java#L400
   private static void configureCompatibilityOptions(JavaCompile compileTask, JavaCompilerConfiguration config) {
+    config.setEnablePreview(compileTask.getOptions().getAllCompilerArgs().stream().anyMatch(it -> it.equals("--enable-preview")));
```
dbmeneses, line 92: `getAllCompilerArgs` calls "Compiler Argument Providers", and one of them, `JavaCompileOptionsForRoom` created by the android runtime, fails because it can't find a file. This provider just adds `-parameters` to the JVM args for annotation processing (collection of runtime metrics)

I've not tried to reproduce (requires setting up android) but this could very well be because the file is deleted after one of the android tasks.

From what I saw, compiler argument providers have a very limited use, and the only use outside tests is `JavaCompileOptionsForRoom` in android. So I'd try to use `getCompilerArgs()` instead of `getAllCompilerArgs` to avoid calling that compiler argument provider.


```
Caused by: java.io.UncheckedIOException: java.io.FileNotFoundException: /tmp/junit11213072525406089847/multi-module-android-studio/app/build/intermediates/annotation_processor_list/release/annotationProcessors.json (No such file or directory)
	at com.android.build.gradle.tasks.JavaCompileUtils.readAnnotationProcessorsFromJsonFile(JavaCompileUtils.kt:272)
	at com.android.build.gradle.tasks.JavaCompileOptionsForRoom.asArguments(JavaCompile.kt:205)
	at org.gradle.api.tasks.compile.CompileOptions.getAllCompilerArgs(CompileOptions.java:329)
```

dbmeneses (approved):

src/main/java/org/sonarqube/gradle/JavaCompilerUtils.java
kaufco: Fixed. CI build does now pass.

PR #166, opened 2023-05-17
NO-JIRA Update README.md file with Integration tests instructions

irina-batinic-sonarsource:


README.md
```diff
@@ -68,6 +68,16 @@ To debug, simply add the parameter:
 
 Now debug remotely by connecting to the port 5005.
 
+### Integration Tests
+By default, Integration Tests are skipped during the build. To run them, you need to follow steps:
```
kaufco, line 72: follow these steps:

README.md
```diff
@@ -68,6 +68,16 @@ To debug, simply add the parameter:
...
+
+First, you need to install the SNAPSHOT version of the root project in the local Maven repository.  
+Then, you need to import the `integationTests` project as a Maven project and ensure that Android SDK is set.
```
kaufco, line 75: I'd suggest to use a bullet list here:
- install the SNAPSHOT ...
- import the `intzegrationTests` ...

kaufco (changes requested):

README.md
ADarko22: also should we mention that it may be required to `export ANDROID_SDK_ROOT`?
For example: `export ANDROID_SDK_ROOT=/usr/lib/android-sdk/` or `export ANDROID_SDK_ROOT=/home/<user>/AndroidSDK`

irina-batinic-sonarsource: Good point, Angelo!
Thank you for your suggestion, both of you. It is updated now.

README.md
```diff
@@ -68,6 +68,16 @@ To debug, simply add the parameter:
 
 Now debug remotely by connecting to the port 5005.
 
+### Integration Tests
+By default, Integration Tests are skipped during the build. To run them, you need to follow these steps:
+
+* Install the SNAPSHOT version of the root project in the local Maven repository.  
+* Import the `integationTests` project as a Maven project and ensure that Android SDK is set.  
+* Set `ANDROID_SDK_HOME` environment variable
```
leveretka, line 76: Shouldn't this be `ANDROID_SDK_ROOT`?

irina-batinic-sonarsource: It seems `ANDROID_SDK_ROOT` is deprecated, `ANDROID_HOME` should be used.
https://developer.android.com/tools/variables#envar

I will update the file.

leveretka: Good catch!

kaufco (approved):

PR #169, opened 2023-05-26
Update version to 4.2-SNAPSHOT

irina-batinic-sonarsource:


leveretka (commented):
I think you should also bump the plugin version used by us, like here:

https://github.com/SonarSource/sonar-scanner-gradle/commit/fecc5c505cceb787344996bde808c6c41bc0c259

kaufco (approved):

PR #171, opened 2023-05-31
Fixed ClassCastException for KotlinJvmProjectExtension and fixed discovery of Spock tests

ADarko22:


ADarko22:
@johann-beleites-sonarsource when merging let's add both of us as co-authors :) we share blame and glory :p

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
```diff
@@ -75,6 +75,8 @@ public class SonarPropertyComputer {
 
   static final String SONAR_SOURCES_PROP = "sonar.sources";
   static final String SONAR_TESTS_PROP = "sonar.tests";
+  public static final String MAIN_SOURCE_SET_SUFFIX = "main";
+  public static final String TEST_SOURCE_SET_SUFFIX = "test";
```
johann-beleites-sonarsource, line 79:
```suggestion
  private static final String MAIN_SOURCE_SET_SUFFIX = "main";
  private static final String TEST_SOURCE_SET_SUFFIX = "test";
```

johann-beleites-sonarsource (approved):
One minor comment, then LGTM!

PR #176, opened 2023-06-01
Update version to 4.3-SNAPSHOT and sonarqube version to 4.2.0.3129

ADarko22:


sonarqube-build.gradle.kts
```diff
@@ -8,7 +8,7 @@ buildscript {
     }
 
     dependencies {
-        classpath("org.sonarsource.scanner.gradle:sonarqube-gradle-plugin:4.1.0.3113")
+        classpath("org.sonarsource.scanner.gradle:sonarqube-gradle-plugin:4.2.0.3127")
```
ADarko22, line 11: wait before merging to clarify with release engineers what version is better to have: `4.2.0.3127` or `4.2.0.3129`

gradle.properties
```diff
@@ -1,5 +1,5 @@
 group=org.sonarsource.scanner.gradle
-version=4.2.1-SNAPSHOT
+version=4.3.0-SNAPSHOT
```
johann-beleites-sonarsource, line 2: Just 4.3 is enough, omit the trailing 0
```suggestion
version=4.3-SNAPSHOT
```

PR #180, opened 2023-07-05
SONARGRADL-125 Add mend task

irina-batinic-sonarsource:


.cirrus.yml
```diff
@@ -88,6 +94,33 @@ build_task:
       path: "**/test-results/**/*.xml"
       format: junit
 
+mend_scan_task:
+  <<: *SETUP_GRADLE_CACHE
+  depends_on:
+    - build
+  # run only on master and long-living branches
+  only_if: $CIRRUS_USER_COLLABORATOR == 'true' && $CIRRUS_TAG == "" && ($CIRRUS_BRANCH == "master" || $CIRRUS_BRANCH =~ "branch-.*")
```
dorian-burihabwa-sonarsource, line 102: Because branches with names  starting with `branch-`are considered long-term branches and are push-protected, it can be difficult to tweak the Mend configuration and test in the future.
I would suggest, we give ourself some easier option by authorizing  branches that start with `mend-`
```suggestion
  only_if: $CIRRUS_USER_COLLABORATOR == 'true' && $CIRRUS_TAG == "" && ($CIRRUS_BRANCH == "master" || $CIRRUS_BRANCH =~ "branch-.*" || $CIRRUS_BRANCH =~ "mend-.*")
```

dorian-burihabwa-sonarsource (changes requested):
LGTM but we should keep an easy way for us to tweak the Mend configuration

PR #182, opened 2023-07-12
SONARGRADL-120 Add .gradle.kts files to the sonar.sources

leveretka:


src/test/groovy/org/sonarqube/gradle/GradleKtsTests.groovy
```diff
@@ -0,0 +1,162 @@
...
+                .build()
+
+        def props = new Properties()
+        props.load(outFile.newDataInputStream())
+
+        then:
+        props["sonar.sources"] == ""
+
+    }
+
```
kaufco, line 130: Maybe add two more test cases:
1. with buildKts, but without settings (neither KTS nor Grooy)
2. without both build and settings

kaufco (changes requested):
Looks nice! Maybe two more test cases.

src/test/groovy/org/sonarqube/gradle/GradleKtsTests.groovy
leveretka: About 1, I can add tests,
About 2 - we can't execute Gradle scanner if it's not a Gradle project, so there should be always build file

kaufco (approved):

PR #185, opened 2023-07-20
SONARGRADL-126 Fix "build.gradle.kts" files of subprojects are indexe…

leveretka:
…d multiple times


jschmid:
Any idea when this will be released?

PR #191, opened 2023-09-19
Normalize paths in tests before performing assertions

leonardo-pilastri-sonarsource:


src/test/groovy/org/sonarqube/gradle/SonarQubePluginTest.groovy
```diff
@@ -725,11 +728,11 @@ class SonarQubePluginTest extends Specification {
     properties["sonar.java.jdkHome"] != null
   }
 
-  def KOTLIN_JVM_SOURCE_FILE =  "src/test/projects/kotlin-jvm-project/src/main/kotlin/Sample.kt"
+  def KOTLIN_JVM_SOURCE_FILE = "src/test/projects/kotlin-jvm-project/src/main/kotlin/Sample.kt"
```
dorian-burihabwa-sonarsource, line 731: Grep'ing around for this constant, it looks like it is not used within the file. We can probably drop it?
```suggestion
```

dorian-burihabwa-sonarsource (changes requested):

dorian-burihabwa-sonarsource (approved):

PR #192, opened 2023-09-19
SONARGRADL-111 Avoid nested paths in sonar.sources/sonar.tests

leonardo-pilastri-sonarsource:


src/test/groovy/org/sonarqube/gradle/AndroidUtilsTest.java
```diff
@@ -128,4 +132,30 @@ void configureForAndroid_androidPropertiesAreSetForAndroidProject_whenVariantDet
     assertEquals(30, resultProperties.get("sonar.android.minsdkversion.max"));
   }
 
+  @Test
+  void configureForAndroid_presenceOfSourceFiles() {
```
dorian-burihabwa-sonarsource, line 136: Let's document and explain why we test sonar.tests and not sonar.sources

dorian-burihabwa-sonarsource: Let's use assertJ for better readibility

src/test/groovy/org/sonarqube/gradle/AndroidUtilsTest.java
```diff
@@ -120,12 +123,22 @@ void configureForAndroid_androidPropertiesAreSetForAndroidProject_whenVariantDet
...
+    when(baseVariant.getSourceSets()).thenReturn(List.of(javaSource));
+
     Map<String, Object> resultProperties = new HashMap<>();
     AndroidUtils.configureForAndroid(project, "myVariant", resultProperties);
 
     assertEquals(true, resultProperties.get("sonar.android.detected"));
     assertEquals(30, resultProperties.get("sonar.android.minsdkversion.min"));
     assertEquals(30, resultProperties.get("sonar.android.minsdkversion.max"));
+    Set<File> sources = (Set<File>) resultProperties.get("sonar.sources");
+    assertTrue(sources.contains(file));
```
dorian-burihabwa-sonarsource, line 141: We may want to test the size of the sources set just to detect regressions in the future.
```suggestion
    assertTrue(sources.size() == 1);
    assertTrue(sources.contains(file));
```

dorian-burihabwa-sonarsource (approved):
LGTM Thanks for the changes. I suggest a minor change to lock down the tests further. Feel free to leave it, if you don't think it adds a value here

PR #195, opened 2023-09-20
SONARGRADL-107 Add integration test with Android Gradle 8.2

dorian-burihabwa-sonarsource:


leveretka (approved):
Overall looks good, let's see if we got any feedback from our Android users.

PR #197, opened 2023-09-22
SONARGRADL-130 Add a property "sonar.skipCompile" to enable/disabe "d…

leveretka:
…ependsOn" on all the compile tasks

### DIFF
--- a/src/test/groovy/org/sonarqube/gradle/SonarQubePluginTest.groovy
+++ b/src/test/groovy/org/sonarqube/gradle/SonarQubePluginTest.groovy
@@ -811,6 +811,47 @@ class SonarQubePluginTest extends Specification {
     properties["sonar.kotlin.gradleProjectRoot"] == rootProject.projectDir as String
   }
 
+  def "Add implicit compile dependencies by default"() {
+    def rootProject = ProjectBuilder.builder().withName("root").build()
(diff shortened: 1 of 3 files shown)


src/test/groovy/org/sonarqube/gradle/SonarQubePluginTest.groovy
```diff
@@ -811,6 +811,47 @@ class SonarQubePluginTest extends Specification {
...
+    def rootProject = ProjectBuilder.builder().withName("root").build()
+    rootProject.pluginManager.apply(JavaPlugin)
+
+    when:
+    rootProject.pluginManager.apply(SonarQubePlugin)
+
+    then:
+    def sonarTask = rootProject.tasks.sonar
+    sonarTask.getDependsOn().size() == 2
+    sonarTask.getMustRunAfter().getDependencies(sonarTask).size() == 1
```
dorian-burihabwa-sonarsource, line 824: Since it should always be the same 3 tasks, should we be explicit in the tasks that are expecting to list?

dorian-burihabwa-sonarsource (approved):
LGTM but it would be even better if we could be explicit in the tasks we expect to find in `dependsOn` and `mustRunAfter`

PR #199, opened 2023-10-02
SONARGRADL-132 Rename parameter `sonar.skipCompile` to `sonar.gradle.skipCompile`

dorian-burihabwa-sonarsource:
Rename the parameter so that its name carries its Gradle specificity. This should make it clearer that this parameter is not meant to be supported by other scanners.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 group=org.sonarsource.scanner.gradle
-version=4.5-SNAPSHOT
+version=4.4.1-SNAPSHOT
 description=Gradle plugin to help analyzing projects with SonarQube
 projectTitle=SonarQube Scanner for Gradle
 kotlinVersion=1.8.20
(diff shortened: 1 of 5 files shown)


gradle.properties
```diff
@@ -1,5 +1,5 @@
 group=org.sonarsource.scanner.gradle
-version=4.5-SNAPSHOT
+version=4.4.1-SNAPSHOT
```
Wohops, line 2: In the release process, they say there is another file to change to prepare release: [integrationTests/pom.xml](https://github.com/SonarSource/sonar-scanner-gradle/blob/master/integrationTests/pom.xml)

Wohops (approved):
One file missing in preparing the new version. I'll take care of it.

Wohops:
PR updated with the version number changed in the missing file.

PR #210, opened 2024-03-18
SONARGRADL-135 Remove usage of deprecated Gradle APIs

dorian-burihabwa-sonarsource:


TWiStErRob:
I can't comment on Jira, so I'll comment here.

@leveretka would you accept a version of this PR that does work on earlier Gradle versions too? When it comes to Gradle compatibility along a long range of versions, it's rarely possible to satisfy with a single code path. In this case a simple `if (GradleVersion.current() > ...) plugins else conventions` would very likely solve your LTS support issues. Later when conventions are removed from Gradle, one of the branches won't compile any more, and at that point the `if` can be removed.

An example of the 2 code paths:
 * https://github.com/bjhargrave/bnd/blob/e26795709c0b9827da61ffe034f2fc6c54cad48f/gradle-plugins/biz.aQute.bnd.gradle/src/main/java/aQute/bnd/gradle/BndPlugin.java#L155-L172

An example of a removal caused by moving to a supported min Gradle version (7.3 / Java 17 similar to yours):
 * https://github.com/bndtools/bnd/pull/5453/files#diff-4082362f5e3eabd13f5d2b584d3503b56cc9f15ffb23b0e8b2ce4dfd0dfdf623

TWiStErRob:
@dorian-burihabwa-sonarsource thoughts?

PR #211, opened 2024-03-18
SONARGRADL-124 Remove implicit dependency of the sonar task on the compile tasks

leveretka:


integrationTests/src/test/java/org/sonarqube/gradle/AndroidTest.java
```diff
@@ -315,11 +335,6 @@ public void noDebugVariant() throws Exception {
         baseDir.resolve("src/main/AndroidManifest.xml"));
     assertThat(Paths.get(props.getProperty("sonar.tests"))).isEqualTo(baseDir.resolve("src/test/java"));
 
-    assertThat(Paths.get(props.getProperty("sonar.java.binaries")))
-      .isEqualTo(baseDir.resolve("build/intermediates/javac/release/classes"));
-    assertThat(Paths.get(props.getProperty("sonar.java.test.binaries")))
-      .isEqualTo(baseDir.resolve("build/intermediates/javac/releaseUnitTest/classes"));
```
dorian-burihabwa-sonarsource, line 321: Let's try and see if we can add compilation so that we can keep these assertions

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿 Let's try and see if we can restore the assertions on the binaries

PR #216, opened 2024-03-21
SONARGRADL-139 Fix ClassCastException when Java compiler args contain…

leveretka:
… non-String values

### DIFF
--- a/src/main/java/org/sonarqube/gradle/JavaCompilerUtils.java
+++ b/src/main/java/org/sonarqube/gradle/JavaCompilerUtils.java
@@ -89,7 +89,8 @@ public class JavaCompilerUtils {
   // Inspired by
   // https://github.com/gradle/gradle/blob/d3e4faca3df507176b67d9b3bb3ee91bf2aa070c/subprojects/language-java/src/main/java/org/gradle/api/tasks/compile/JavaCompile.java#L400
   private static void configureCompatibilityOptions(JavaCompile compileTask, JavaCompilerConfiguration config) {
-    config.setEnablePreview(compileTask.getOptions().getCompilerArgs().stream().anyMatch(it -> it.equals("--enable-preview")));
+    List<?> compilerArgs = compileTask.getOptions().getCompilerArgs();
+    config.setEnablePreview(compilerArgs.stream().anyMatch(it -> it.equals("--enable-preview")));
     if (areToolchainsSupported() && ToolchainUtils.hasToolchains(compileTask)) {
       ToolchainUtils.configureCompatibilityOptions(compileTask, config);
     } else {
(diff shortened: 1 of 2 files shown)


src/main/java/org/sonarqube/gradle/JavaCompilerUtils.java
```diff
@@ -89,7 +89,8 @@ private static boolean areToolchainsSupported() {
   // Inspired by
   // https://github.com/gradle/gradle/blob/d3e4faca3df507176b67d9b3bb3ee91bf2aa070c/subprojects/language-java/src/main/java/org/gradle/api/tasks/compile/JavaCompile.java#L400
   private static void configureCompatibilityOptions(JavaCompile compileTask, JavaCompilerConfiguration config) {
-    config.setEnablePreview(compileTask.getOptions().getCompilerArgs().stream().anyMatch(it -> it.equals("--enable-preview")));
+    List<?> compilerArgs = compileTask.getOptions().getCompilerArgs();
+    config.setEnablePreview(compilerArgs.stream().anyMatch(it -> it.equals("--enable-preview")));
```
dorian-burihabwa-sonarsource, line 93: Do we know how breaking down into multiple expressions is safer? I am not sure I understand how this change helps 🤔

dorian-burihabwa-sonarsource (changes requested):
Thank you for adding an example of the type of cases we are supposed to handle. But I am not sure why looking at the code change that we will be able to tolerate these changes better

src/main/java/org/sonarqube/gradle/JavaCompilerUtils.java
leveretka: This is the generics and streams issue.  The compile-time type of `compileTask.getOptions().getCompilerArgs()` is List<String>, so `anyMatch` expects a `Predicate<String>`. During runtime all types are erased, so to ensure the type is correct there's an implicit cast to `String`.

Due to the fact that the input we receive from Groovy scripts, there could be any Objects. For example `File`. That was the issue when we tried to cast `File` to `String` and failed.

With my change, the compile-time type of the `compilerArgs` variable is List<Object>, but I can't use `List<Object>` directly because in Java generics are invariant. But with this change, we're calling the stream from `List<Object>` and the `anyMatch` method expects `Predicate<Object>`. Then no cast is performed and no exception is thrown.

src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
```diff
@@ -280,12 +280,12 @@ class FunctionalTests extends Specification {
...
-          .forwardOutput()
-          .withArguments('sonarqube', '-Dsonar.scanner.dumpToFile=' + outFile.toAbsolutePath())
-          .withPluginClasspath()
-          .build()
+                .withGradleVersion(gradleVersion)
+                .withProjectDir(testProjectDir.toFile())
+                .forwardOutput()
+                .withArguments('sonarqube', '-Dsonar.scanner.dumpToFile=' + outFile.toAbsolutePath())
+                .withPluginClasspath()
+                .build()
```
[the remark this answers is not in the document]
leveretka: Probably I have different formatting options in 2 different machines, will revert

dorian-burihabwa-sonarsource (approved):

PR #227, opened 2024-06-19
SCANGRADLE-145 The scan all behavior excludes external report files

dorian-burihabwa-sonarsource:


src/main/java/org/sonarqube/gradle/SonarUtils.java
```diff
@@ -38,6 +40,11 @@
 
 public class SonarUtils {
 
+  private static final Pattern COVERAGE_REPORT_PATH_PROPERTY_PATTERN = Pattern.compile(
```
ADarko22, line 43: In the scanner for maven we use the pattern: `"^sonar\\..*[rR]eportPaths?$"`. Should we try to be consistent?
Does the pattern need to be necessarily  `CASE_INSENSITIVE`?

ADarko22 (changes requested):
Great job! Just two questions that may improve readibility and align with the scanner for Maven. Then I would be okay for merging :)

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
```diff
@@ -217,7 +219,16 @@ private static void computeScanAllProperties(Project project, Map<String, Object
       .map(File::toPath)
       .collect(Collectors.toSet());
 
-    SourceCollector visitor = new SourceCollector(allModulesExistingSources, skippedDirs, Set.of(), false);
+    Path root = Path.of(findProjectBaseDir(properties));
+    Set<Path> excludedFiles = extractReportPaths(properties).stream()
```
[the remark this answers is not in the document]
dorian-burihabwa-sonarsource: We could do it but it makes testing the logic a little more complicated: we need to have a `sonar.projectBaseDir` defined in all tests otherwise the code crashes because it does not seem to be built to work around the missing variable.
I will see what I can do

src/main/java/org/sonarqube/gradle/SonarUtils.java
dorian-burihabwa-sonarsource: I intentionally made the pattern more complex to cover both external reports and [coverage reports](https://docs.sonarsource.com/sonarqube/latest/analyzing-source-code/test-coverage/test-coverage-parameters/). WDYT?

dorian-burihabwa-sonarsource: The case insensitive part is not absolutely necessary but I felt like it meant the pattern more readable. WDYT?

ADarko22: Actually, this is a good idea and it would be nice to have it in the Scanner for Maven as well :)
Do you think that having 2 regexes would be more readable, or is it clear enough by having a more attentive look?

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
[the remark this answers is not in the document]
ADarko22: If testing doesn't turn out to be too painful I would prefer having the logic in the SonarUtils, for readability and consistency

ADarko22 (approved):
LGTM!

PR #228, opened 2024-06-20
SCANGRADLE-148 Gradle scanner scans some files in hidden folders

ADarko22:
This changes are on top of https://github.com/SonarSource/sonar-scanner-gradle/pull/226 that should be merged first

### DIFF
--- a/integrationTests/src/test/java/org/sonarqube/gradle/GradleTest.java
+++ b/integrationTests/src/test/java/org/sonarqube/gradle/GradleTest.java
@@ -355,11 +355,13 @@ public class GradleTest extends AbstractGradleIT {
         baseDir.resolve("gradlew").toString(),
         baseDir.resolve("gradlew.bat").toString(),
         baseDir.resolve("gradle/wrapper/gradle-wrapper.properties").toString(),
-        baseDir.resolve(".gradle/buildOutputCleanup/cache.properties").toString()
+        baseDir.resolve(".hiddenDir/file.txt").toString(),
+        baseDir.resolve(".hiddenConfig").toString()
       ).doesNotContain(
         baseDir.resolve("skippedModule/build.gradle.kts").toString(),
         baseDir.resolve("skippedModule/skippedSubmodule/skippedSubmoduleScript.sh").toString(),
-        baseDir.resolve("skippedModule/skippedSubmodule/build.gradle.kts").toString()
+        baseDir.resolve("skippedModule/skippedSubmodule/build.gradle.kts").toString(),
+        baseDir.resolve(".hiddenDir/script.py").toString()
       );
 
     assertThat(props.getProperty(":module.sonar.sources")).isEqualTo(baseDir.resolve("module/src/main/java").toString());
--- a/integrationTests/src/test/java/org/sonarqube/gradle/GradleTest.java
+++ b/integrationTests/src/test/java/org/sonarqube/gradle/GradleTest.java
@@ -354,12 +354,12 @@ public class GradleTest extends AbstractGradleIT {
         baseDir.resolve("module/submodule/submoduleScript.sh").toString(),
         baseDir.resolve("gradlew").toString(),
         baseDir.resolve("gradlew.bat").toString(),
-        // Will be fixed by SCANGRADLE-149
-        baseDir.resolve("skippedModule/skippedSubmodule/build.gradle.kts").toString(),
         baseDir.resolve("gradle/wrapper/gradle-wrapper.properties").toString()
       ).doesNotContain(
         baseDir.resolve("skippedModule/build.gradle.kts").toString(),
-        baseDir.resolve("skippedModule/skippedSubmodule/skippedSubmoduleScript.sh").toString());
+        baseDir.resolve("skippedModule/skippedSubmodule/skippedSubmoduleScript.sh").toString(),
+        baseDir.resolve("skippedModule/skippedSubmodule/build.gradle.kts").toString()
+      );
 
     assertThat(props.getProperty(":module.sonar.sources")).isEqualTo(baseDir.resolve("module/src/main/java").toString());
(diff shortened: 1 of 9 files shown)


src/main/java/org/sonarqube/gradle/SourceCollector.java
```diff
@@ -101,19 +112,27 @@ public SourceCollector(Set<Path> existingSources, Set<Path> directoriesToIgnore,
...
+      if (relativePath.toFile().isHidden()) {
+        return true;
+      }
+      relativePath = relativePath.getParent();
+    }
+    return false;
+  }
+
+  private boolean isChildOrGrandChildOfRoot(Path path) {
+    return path.getParent() != null && (path.getParent().equals(root) || path.getParent().getParent().equals(root));
```
alban-auzeill, line 135: `getParent().getParent()` may be null
```java
    Path root = Path.of("a");
    Path path = Path.of("b/c");
    path.getParent().getParent().equals(root); // NullPointerException
 ```
You can flip equals `root.equals(path.getParent().getParent())`

File: src/main/java/org/sonarqube/gradle/SourceCollector.java [hunk 2]
```diff
@@ -92,7 +98,12 @@ public Set<Path> getCollectedSources() {
 
   private final Set<Path> collectedSources = new HashSet<>();
 
-  public SourceCollector(Set<Path> existingSources, Set<Path> directoriesToIgnore, Set<Path> excludedFiles, boolean shouldCollectJavaAndKotlinSources) {
+  public static Builder builder() {
+    return new Builder();
+  }
+
+  public SourceCollector(Path root, Set<Path> existingSources, Set<Path> directoriesToIgnore, Set<Path> excludedFiles, boolean shouldCollectJavaAndKotlinSources) {
```
dorian-burihabwa-sonarsource, line 105: If we create an object using the builder pattern, is there a reason to keep the constructor public?

integrationTests/src/test/java/org/sonarqube/gradle/GradleTest.java
```diff
@@ -354,7 +354,8 @@ public void testScanAllOnMultiModuleWithSubModulesProjectCollectsTheExpectedSour
         baseDir.resolve("module/submodule/submoduleScript.sh").toString(),
         baseDir.resolve("gradlew").toString(),
         baseDir.resolve("gradlew.bat").toString(),
-        baseDir.resolve("gradle/wrapper/gradle-wrapper.properties").toString()
+        baseDir.resolve("gradle/wrapper/gradle-wrapper.properties").toString(),
+        baseDir.resolve(".gradle/buildOutputCleanup/cache.properties").toString()
```
dorian-burihabwa-sonarsource, line 358: What is the motivation for analyzing this `buildOutputCleanup` directory? Is it a static, committed file or is an artifact introduced by gradle or on the task that ran before the sonar task?

dorian-burihabwa-sonarsource (changes requested):
There are a few things we can tweak for clarity and consistency. I am not super comfortable with the large inclusion of hidden files and I am worried about the behavior of the scanner with folders such as scm folders or virtual environments

src/main/java/org/sonarqube/gradle/SourceCollector.java
```diff
@@ -101,19 +112,27 @@ public SourceCollector(Set<Path> existingSources, Set<Path> directoriesToIgnore,
...
     return FileVisitResult.CONTINUE;
   }
 
-  private static boolean isHidden(Path path) {
-    return StreamSupport.stream(path.spliterator(), true)
-      .anyMatch(token -> token.toString().startsWith("."));
+  private boolean isHidden(Path path) {
+    Path relativePath = root.relativize(path);
+    while (relativePath != null && !relativePath.equals(root)) {
+      if (relativePath.toFile().isHidden()) {
```
alban-auzeill, line 126: No need to convert the path to call `relativePath.toFile().isHidden()` you can do it using `Files.isHidden(path)`
But why changing `.startsWith(".")` by `.isHidden(path)`?
I don't know if it is reliable. If I extract a GitHub repo on a Linux and a Windows machine, will both OSs consider the same files hidden?

alban-auzeill (changes requested):

src/main/java/org/sonarqube/gradle/SourceCollector.java
```diff
@@ -81,6 +80,13 @@ public class SourceCollector implements FileVisitor<Path> {
     ".kt")).map(ext -> ext.toLowerCase(Locale.ROOT))
     .collect(Collectors.toSet());
 
+  private static final Set<String> INCLUDE_EXTENSIONS_FOR_HIDDEN_FILES = Set.of(".env", ".json", ".yml", ".yaml", ".properties", ".db", ".srcMainJava", ".py", ".xml", ".sh",
+    ".bash", ".ksh", ".zsh", ".bat", ".ps1", ".txt", ".config", ".settings", ".cnf");
```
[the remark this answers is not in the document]
dorian-burihabwa-sonarsource: Let's check if `srcMainJava` is part of the inclusions we want to add

integrationTests/src/test/java/org/sonarqube/gradle/GradleTest.java
ADarko22: As discussed we'll exclude visiting the `.gradle` and `.git` directories

integrationTests/src/test/resources/multi-module-with-submodules/build.gradle.kts
```diff
@@ -1,6 +1,6 @@
 plugins {
   java
-  id("org.sonarqube") version "${version}"
+  id("org.sonarqube") version "5.1-SNAPSHOT"
```
alban-auzeill, line 3: Oops. we need to revert this change!

dorian-burihabwa-sonarsource (approved):

PR #230, opened 2024-06-24
SCANGRADLE-143 The scanner for Gradle avoids symbolic links when collecting files outside of conventional sonar.sources

alban-auzeill:


dorian-burihabwa-sonarsource (approved):
Thank you for the investigation. LGTM 👍🏿

PR #231, opened 2024-06-26
SCANGRADLE-152 Additional files added by "scanAll" should be categorized as TEST when the path looks like a test

alban-auzeill:


src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
```diff
@@ -85,6 +87,22 @@ public class SonarPropertyComputer {
...
+   * But exclude not test related English words.
+   */
+  private static final Pattern TEST_FILE_PATH_PATTERN = Pattern.compile(
+    // Exclude valid English words ending with "test": contest, protest, detest, attest
+    "(?<!con|pro|de|at)" +
+      // Find path containing "test"
+      "test" +
+      // Exclude valid English words starting with "test":
+      // - testate, testator, testatrix, testament, testimonial, testimony, testiness, testy
+      "(?!ate|ator|atrix|ament|imonial|imony|iness|y)",
```
dorian-burihabwa-sonarsource, line 103: Let's add the source or method used to generate this list of words so that we can make sure in the future that the list is up to date or needs to be challenged

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
```diff
@@ -244,7 +272,17 @@ private static void computeScanAllProperties(Project project, Map<String, Object
...
+
+  public enum InputFileType {
+    MAIN, TEST
+  }
+
+  // VisibleForTesting
+  static InputFileType findProjectFileType(Path projectDir, Path filePath) {
+    String relativePath = projectDir.relativize(filePath).toString();
+    return TEST_FILE_PATH_PATTERN.matcher(relativePath).find() ? InputFileType.TEST : InputFileType.MAIN;
   }
```
dorian-burihabwa-sonarsource, line 286: I think it would make sense to move this logic to the SonarUtils class. WDYT?

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
```diff
@@ -244,7 +272,17 @@ private static void computeScanAllProperties(Project project, Map<String, Object
       .sorted()
       .collect(Collectors.toList());
 
-    properties.put(ScanProperties.PROJECT_SOURCE_DIRS, SonarUtils.joinAsCsv(mergedSources));
+    properties.put(sourcePropertyToUpdate, SonarUtils.joinAsCsv(mergedSources));
+  }
+
+  public enum InputFileType {
+    MAIN, TEST
+  }
```
dorian-burihabwa-sonarsource, line 280: Can we use the types from sonar-plugin-api in this part of the code?

src/test/groovy/org/sonarqube/gradle/GradleKtsTests.groovy
```diff
@@ -83,7 +83,7 @@ class GradleKtsTests extends Specification {
 
         then:
         props["sonar.sources"] == buildFile.toRealPath().toString()
-
+        props["sonar.tests"] == null
```
dorian-burihabwa-sonarsource, line 86: Question: Can we use `props.containsKey("sonar.tests")` in this code?

src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
```diff
@@ -42,6 +42,9 @@ class FunctionalTests extends Specification {
     def setup() {
         settingsFile = testProjectDir.resolve('settings.gradle')
         buildFile = testProjectDir.resolve('build.gradle')
+        testProjectDir.resolve('integrationTests').toFile().mkdir()
```
dorian-burihabwa-sonarsource, line 45: Question: Would it make sense to rename it to "projectDir"?

src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
```diff
@@ -42,6 +42,9 @@ class FunctionalTests extends Specification {
...
+        testProjectDir.resolve('integrationTests').resolve("run-all.sh") << "# a script file"
+        testProjectDir.resolve('test-license.sh') << "# a script file"
```
dorian-burihabwa-sonarsource, line 47: To make it a little more explicit
```suggestion
        testProjectDir.resolve('integrationTests').resolve("run-all.sh") << "# a test script file"
        testProjectDir.resolve('test-license.sh') << "# a test script file"
```

src/test/java/org/sonarqube/gradle/SonarPropertyComputerTest.java
```diff
@@ -0,0 +1,49 @@
...
+    assertThat(findProjectFileType(projectDir, Paths.get(project, "script/test/run.sh"))).isEqualTo(InputFileType.TEST);
+    assertThat(findProjectFileType(projectDir, Paths.get(project, "script/integrationTests/run.sh"))).isEqualTo(InputFileType.TEST);
+    assertThat(findProjectFileType(projectDir, Paths.get(project, "script/run-allTests.sh"))).isEqualTo(InputFileType.TEST);
+    assertThat(findProjectFileType(projectDir, Paths.get(project, "script/runContest.sh"))).isEqualTo(InputFileType.MAIN);
+    assertThat(findProjectFileType(projectDir, Paths.get(project, "protest-for-freedom.json"))).isEqualTo(InputFileType.MAIN);
+    assertThat(findProjectFileType(projectDir, Paths.get(project, "detest-bad-code.md"))).isEqualTo(InputFileType.MAIN);
+    assertThat(findProjectFileType(projectDir, Paths.get(project, "testate/write-testament-from-testator-and-testatrix.sh"))).isEqualTo(InputFileType.MAIN);
+    assertThat(findProjectFileType(projectDir, Paths.get(project, "generate-testimonial.sh"))).isEqualTo(InputFileType.MAIN);
+    assertThat(findProjectFileType(projectDir, Paths.get(project, "run-testy-testiness-testimony.sh"))).isEqualTo(InputFileType.MAIN);
+  }
```
dorian-burihabwa-sonarsource, line 47: For readability purposes, I think it would make sense to split the test cases between those that are supposed to test and order them in a consistent way. This would allow us to spot and add regression test cases in the future

dorian-burihabwa-sonarsource (approved):
Thank you for the investigation and quick turn-around. I have a couple of suggestions:

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
alban-auzeill: InputFile.Type is not available

src/test/groovy/org/sonarqube/gradle/SonarQubePluginTest.groovy
```diff
@@ -936,6 +943,18 @@ class SonarQubePluginTest extends Specification {
     return Paths.get(pathString).normalize().toAbsolutePath();
   }
 
+  private String pathList(Project project, String propertyName) {
```
dorian-burihabwa-sonarsource, line 946: We should push this utility method further down in the file, away from the tests and closer to the other utility methods.

src/test/groovy/org/sonarqube/gradle/SonarQubePluginTest.groovy
```diff
@@ -999,36 +1018,36 @@ class SonarQubePluginTest extends Specification {
...
+      build.gradle.kts
+      module1/build.gradle.kts
+      module1/extras/pyScriptM1.py
+      module1/scriptM1.sh
+      settings.gradle.kts
+      """.stripIndent().trim()
+
+    testSources == """
+      .hidden/folder/test-config.config
+      """.stripIndent().trim()
```
dorian-burihabwa-sonarsource, line 1043: Can we simplify this?
```suggestion
    testSources == ".hidden/folder/test-config.config"
```

src/test/groovy/org/sonarqube/gradle/SonarQubePluginTest.groovy
```diff
@@ -999,36 +1018,36 @@ class SonarQubePluginTest extends Specification {
...
+
+    module1Sources == """
+      module1/src/main/java
+      """.stripIndent().trim()
```
dorian-burihabwa-sonarsource, line 1047: Can we simplify this?
```suggestion
    module1Sources == "module1/src/main/java"
```

dorian-burihabwa-sonarsource (changes requested):

src/test/groovy/org/sonarqube/gradle/SonarQubePluginTest.groovy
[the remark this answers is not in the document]
alban-auzeill: I don't know how to create a nice, readable test failure report using this Groovy test context when the argument is a `List`. By returning a `String`, the failure report becomes easier to read and supported by the IDE diff tool.

PR #234, opened 2024-07-02
SCANGRADLE-152 Fix issue with "sonar.tests" in child modules

ADarko22:
Child modules copy the "sonar.tests" property from the parent when it is not defined. This behavior is encoded in the scanner engine, and it breaks the logic defined in [SCANGRADLE-152](https://sonarsource.atlassian.net/browse/SCANGRADLE-152) causing doubled indexing.

[SCANGRADLE-152]: https://sonarsource.atlassian.net/browse/SCANGRADLE-152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
+++ b/src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
@@ -136,8 +136,9 @@ public class SonarPropertyComputer {
 
     overrideWithUserDefinedProperties(project, rawProperties);
 
-    // This is required if "sonar.sources" are neither found nor defined by user
+    // This is required if "sonar.sources" or "sonar.tests" are neither found nor defined by user; possible undesired behaviors due to the copy of those properties from the parent
     rawProperties.putIfAbsent(ScanProperties.PROJECT_SOURCE_DIRS, "");
+    rawProperties.putIfAbsent(ScanProperties.PROJECT_TEST_DIRS, "");
 
     if (project.equals(targetProject)) {
       rawProperties.putIfAbsent("sonar.projectKey", computeProjectKey());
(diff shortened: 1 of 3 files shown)


src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
```diff
@@ -136,8 +136,9 @@ private void computeDefaultProperties(Project project, Map<String, Object> prope
 
     overrideWithUserDefinedProperties(project, rawProperties);
 
-    // This is required if "sonar.sources" are neither found nor defined by user
+    // This is required if "sonar.sources" or "sonar.tests" are neither found nor defined by user; possible undesired behaviors due to the copy of those properties from the parent
     rawProperties.putIfAbsent(ScanProperties.PROJECT_SOURCE_DIRS, "");
+    rawProperties.putIfAbsent(ScanProperties.PROJECT_TEST_DIRS, "");
```
dorian-burihabwa-sonarsource, line 141: Let's add a more specific command here
```suggestion
    // These empty assignments are required because modules with no `sonar.sources` or `sonar.tests` value inherit the value from their parent module.
    // This can eventually lead to a double indexing issue in the scanner-engine.
    rawProperties.putIfAbsent(ScanProperties.PROJECT_SOURCE_DIRS, "");
    rawProperties.putIfAbsent(ScanProperties.PROJECT_TEST_DIRS, "");
```

dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿 Let's apply the comment change to make it easier on ourselves next time

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
dorian-burihabwa-sonarsource: We can link to this [ticket](https://sonarsource.atlassian.net/browse/SCANGRADLE-151) and add a TODO comment.

PR #237, opened 2024-07-03
Remove conditions from manually-triggered releasability check

johann-beleites-sonarsource:
No need for conditions, as we only trigger this manually.


dorian-burihabwa-sonarsource (approved):
LGTM 👍🏿

ADarko22 (approved):

PR #242, opened 2024-07-12
Lazily register sonar and sonarqube tasks

GreyTeardrop:
Replace eager task registration via `TaskContainer.create()` with lazy registration via `TaskContainer.register()`.

Motivation: analyzing a build scan, I've noticed that my build had 2 tasks created immediately. Those turned out to be  `sonar` and `sonarqube`.

CleanShot 2024-07-13 at 00 31 05@2x

It would be great to avoid creating those tasks if SonarScanner isn't executed.


pwagland (approved):
Changes look good to me, and saves small amount of time per project. When you have hundreds or thousands to project these small bits add up. The tasks are correctly configured lazily.

dorian-burihabwa-sonarsource:
We need to check that the API is compatible with the lowest version of Gradle we support 5.6.4

GabrielFleischer:
A ticket has been created, it will be taken care of in a future sprint. 

https://sonarsource.atlassian.net/browse/SCANGRADLE-158

GreyTeardrop:
I see, thank you!

PR #247, opened 2024-09-10
Make value of SonarProperties#property nullable

Vampire:
I'm porting a Groovy DSL build to Kotlin DSL.

Before I did things like
```groovy
property('foo', [...] ?: null)
```
to override `foo` and either set it to some value or unset it otherwise.

In Kotlin DSL you cannot do
```kotlin
property("foo", listOf(...).takeIf { it.isNotEmpty() })
```
currently due to the `@ParametersAreNonnullByDefault` in the `package-info.java` and Kotlin respecting that, so having the value parameter non-nullable.
Work-around is to use
```kotlin
properties["foo"] = listOf(...).takeIf { it.isNotEmpty() }
```
or
```kotlin
listOf(...).also { if (it.isNotEmpty()) property("foo", it) else properties.remove("foo") }
``` 

This PR makes the value parameter nullable, so you can also use the first Kotlin snippet above.

### DIFF
--- a/src/main/java/org/sonarqube/gradle/SonarProperties.java
+++ b/src/main/java/org/sonarqube/gradle/SonarProperties.java
@@ -19,6 +19,7 @@
  */
 package org.sonarqube.gradle;
 
+import javax.annotation.Nullable;
 import java.util.Map;
 
 /**
@@ -45,7 +46,7 @@ public class SonarProperties {
    * @param key the key of the property to be added
    * @param value the value of the property to be added
    */
-  public void property(String key, Object value) {
+  public void property(String key, @Nullable Object value) {
     properties.put(key, value);
   }
 


dorian-burihabwa-sonarsource:
hey @Vampire, thank you for the suggestion. Do you have any tests that effectively show that a nullable value actually removes the property from the collection?

Vampire:
I don't really get your comment, sorry. 

This is not a suggestion, but a definite bug report.

And it of course does not remove the property from the collection, why should it? It sets is value in the collection to `null`.

dorian-burihabwa-sonarsource:
> I don't really get your comment, sorry.
> 
> This is not a suggestion, but a definite bug report.

My bad, I did not understand what you were trying to achieve here.

> And it of course does not remove the property from the collection, why should it? It sets is value in the collection to `null`.

Could you share, here or in the community forums, an example of a build configuration that triggers the bug?

PR #248, opened 2024-09-10
More reliably detect own version

Vampire:
I modified a build and checked what changes happened to the generated SonarQube properties when using `sonar.scanner.dumpToFile`.
During this I stumbled upon `sonar.scanner.appVersion` being set to `0.60.3/Gradle 7.6.4` now.
The reasons is the very generically named and pathed `/version.txt`.
I now also have the https://github.com/Splitties/refreshVersions/ plugin in the classpath and it comes before this plugin.
That plugin happens to also have a `version.txt` file in the root of the classpath with its version inside.
This plugin then found that file and thus identified itself as version `0.60.3` wrongly.

This PR renames the `version.txt` file to a more unique name that is unlikely to clash with some other project
to more reliably detect the own version.

An alternative would be to put the file into a directory under the namespace of this plugin like `org/sonarqube/gradle/version.txt`.


dorian-burihabwa-sonarsource:
Hey @Vampire,
I created a ticket to track the issue [here](https://sonarsource.atlassian.net/browse/SCANGRADLE-174)

Vampire:
Will this PR ever considered? :-)

dorian-burihabwa-sonarsource:
Hey @Vampire, yes we will be setting some time aside this week and the next to reveiw the open PRs on the project

alban-auzeill:
@Vampire, I decided to apply on your PR, what you have suggested in the above comment, to move the version file to the `org.sonarqube.gradle` package. So I'm closing this PR and will continue on the one I can modify: #296

Vampire:
Hm, sorry, usually I have the "Maintainers can edit the PR" checkbox enabled, dunno why not here. Thanks for bringing it over the goalline. :-)

PR #249, opened 2024-09-20
[SCANGRADLE-156] Set POM information on all publications

Vampire:


Vampire:
It was maybe stale, but not caused by me, please reopen unless it was already fixed.

PR #258, opened 2024-11-05
SCANGRADLE-178 increase android sdk to be compatible with android gradle 8.3.0

erwan-serandour:
[SCANGRADLE-178](https://sonarsource.atlassian.net/browse/SCANGRADLE-178)

The SonarSource Team

[SCANGRADLE-178]: https://sonarsource.atlassian.net/browse/SCANGRADLE-178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -122,8 +122,8 @@ gradle_qa_task:
       - GRADLE_VERSION: 8.3
       - GRADLE_VERSION: 8.0
       - GRADLE_VERSION: 7.6.4
-#- GRADLE_VERSION: 7.5.1
-#        ANDROID_GRADLE_VERSION: 7.1.0
+      - GRADLE_VERSION: 7.5.1
+        ANDROID_GRADLE_VERSION: 7.1.0
   script:
     - ./cirrus/cirrus-qa.sh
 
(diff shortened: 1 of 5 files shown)


.cirrus.yml
```diff
@@ -122,8 +122,8 @@ gradle_qa_task:
       - GRADLE_VERSION: 8.3
       - GRADLE_VERSION: 8.0
       - GRADLE_VERSION: 7.6.4
-#- GRADLE_VERSION: 7.5.1
-#        ANDROID_GRADLE_VERSION: 7.1.0
+      - GRADLE_VERSION: 7.5.1
+        ANDROID_GRADLE_VERSION: 7.1.0
```
dorian-burihabwa-sonarsource, line 126: Consider re-enabling that configuration as part of a separate commit

dorian-burihabwa-sonarsource (changes requested):
LGTM, but the title of the PR needs to be changed to match the ticket it refers to because as the current proposal does not test with AGP 8.3.0.
I think the PR should be renamed and the commits split in two parts with the first bumping the SDK version of the test projects and the second re-enabling the tests with AGP 7.1.0 and Gradle 7.5.1 (both commits can refer to the same ticket).

.cirrus.yml
erwan-serandour: I think it is a good idea, I thought about splitting but was a bit too lazy. :+1:

PR #294, opened 2025-04-30
SCANGRADLE-155 Log Sonar plugin version and execution context information

alban-auzeill:


src/main/java/org/sonarqube/gradle/SonarTask.java
```diff
@@ -149,8 +148,26 @@ public void run() {
...
-  private String getPluginVersion() {
-    InputStream inputStream = this.getClass().getResourceAsStream("/version.txt");
+  private static void logEnvironmentInformation() {
+    if (LOGGER.isInfoEnabled()) {
+      LOGGER.info("org.sonarqube Gradle plugin {}", getPluginVersion());
+      LOGGER.info("Java {} {} ({}-bit)",
+        System.getProperty("java.version"),
+        System.getProperty("java.vm.vendor"),
+        System.getProperty("sun.arch.data.model"));
+      LOGGER.info( "{} {} ({})",
```
dorian-burihabwa-sonarsource, line 158:
```suggestion
      LOGGER.info("{} {} ({})",
```

src/main/java/org/sonarqube/gradle/SonarTask.java
```diff
@@ -127,6 +125,7 @@ public void run() {
     if (isSkippedWithProperty(mapProperties)) {
       return;
     }
+    logEnvironmentInformation();
```
dorian-burihabwa-sonarsource, line 128: I feel like this information should come up as the first information about the task. I don't think it should show up after the information about using the deprecated `sonarqube` task

dorian-burihabwa-sonarsource: Ideally, it would be the first thing called in the task runs

src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
```diff
@@ -233,6 +233,64 @@ class FunctionalTests extends Specification {
...
+        nonNull(
+          assertThat(result.output)
+            .containsPattern('org.sonarqube Gradle plugin \\d+\\.\\d+')
+            .containsPattern('Java \\d+')
+            .contains('(64-bit)')
+            .contains('GRADLE_OPTS=-Dfoo=bar')
+        )
+    }
+
+    def "does not log execution context when sonar.skip is true"() {
```
dorian-burihabwa-sonarsource, line 267: I think we should still print the message even if the project/subproject is skipped

dorian-burihabwa-sonarsource (changes requested):
To be consistent with the [maven scanner](https://github.com/SonarSource/sonar-scanner-maven/blob/ae454f5994a48ec3a1381adf63609e6b0e437747/sonar-maven-plugin/src/main/java/org/sonarsource/scanner/maven/bootstrap/ScannerBootstrapper.java#L71), log the environment info should be the first thing done

dorian-burihabwa-sonarsource (approved):

PR #304, opened 2025-06-18
SCANGRADLE-234 Add Gradle 9.0.0-rc-1 to ITs

dorian-burihabwa-sonarsource:
[SCANGRADLE-234](https://sonarsource.atlassian.net/browse/SCANGRADLE-234)

[SCANGRADLE-234]: https://sonarsource.atlassian.net/browse/SCANGRADLE-234?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -147,6 +147,7 @@ gradle_qa_task:
     memory: 4G
   env:
     matrix:
+      - GRADLE_VERSION: 9.0.0-rc-1
       - GRADLE_VERSION: 8.13
       - GRADLE_VERSION: 8.10
       - GRADLE_VERSION: 8.4
--- a/integrationTests/src/test/resources/AndroidTestingBlueprintWithDynamicFeatureModule/module-plain-java/build.gradle
+++ b/integrationTests/src/test/resources/AndroidTestingBlueprintWithDynamicFeatureModule/module-plain-java/build.gradle
@@ -4,8 +4,8 @@ plugins {
 }
 
 // In case Java 8+ is the default JDK, force Java 7 as this module will be used from Android.
-sourceCompatibility = 1.8
-targetCompatibility = 1.8
+sourceCompatibility = JavaVersion.1_8
+targetCompatibility = JavaVersion.1_8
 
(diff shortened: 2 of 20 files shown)


ben-manes:
FYI 9.0.0-rc-1 it fails due to cross-project resolutions being disallowed (see [buildscan](https://caffeine.gradle-enterprise.cloud/s/qtxhvrdkz5kag/failures?expanded-stacktrace=WyIwLTEiXQ)).

ben-manes:
fyi, debugging [discussions](https://github.com/ben-manes/gradle-versions-plugin/issues/948#issuecomment-3005970612) wrt gradle-version-plugin and the `--no-parallel` trick might work for you too as a stop-gap.

emartynov:
Gradle 9 RC2 is out. Do we have a solution for this?

G00fY2:
So 9.0.0-rc-3 was released which contains https://github.com/gradle/gradle/issues/34190.

Now the error about `Resolution of the configuration ':app:debugCompileClasspath' was attempted without an exclusive lock` is gone. But instead we get:

```
> Task :sonar FAILED
1 problem was found storing the configuration cache.
- Task `:sonar` of type `org.sonarqube.gradle.SonarTask`: error writing value of type 'org.gradle.api.internal.provider.DefaultMapProperty'
```

Workaround with `--no-parallel` still works.

@ben-manes do you have any ideas? Is this a new issue or still somewhat related to https://github.com/gradle/gradle/issues/34064?

**Edit:** After adding `--stacktrace` flag I still got the same error like initially with 9.0.0-rc-1.

ben-manes:
I haven’t tried without —no-parallel. You can reach out to the gradle folks 

I know often my build would capture a provider too early in the chain for the configuration cache and I needed to adjust it to the last provider to not capture configurations, etc. it may be here that provider is too raw, but the snippet lacks the stack trace to know what the input is.

PR #311, opened 2025-08-19
SCANGRADLE-243 Add Gradle 9.0.0 to ITs

alban-auzeill:
[SCANGRADLE-243](https://sonarsource.atlassian.net/browse/SCANGRADLE-243)

[SCANGRADLE-243]: https://sonarsource.atlassian.net/browse/SCANGRADLE-243?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/integrationTests/src/test/java/org/sonarqube/gradle/AbstractGradleIT.java
+++ b/integrationTests/src/test/java/org/sonarqube/gradle/AbstractGradleIT.java
@@ -139,11 +139,6 @@ public abstract class AbstractGradleIT {
         value = dependenciesInGradleCache.matcher(value).replaceAll(res ->
         // Java 11 does not support res.group("groupId") this is why there is res.group(1)
         "\\${M2}/repository/" + res.group(1).replace(".", "/") + "/${artifactId}/${version}/${artifactId}-${version}.jar");
-        // In the Gradle configuration there are some FileCollection that have a random order, like the transitive dependencies in "sonar.java.test.libraries"
-        // so we are forced to sort the entries to have a reproducible result
-        String[] parts = value.split(",", -1);
-        Arrays.sort(parts);
-        value = String.join(",", parts);
         result.put(key, value);
       });
     return result;
--- a/integrationTests/src/test/java/org/sonarqube/gradle/Gradle9Test.java
+++ b/integrationTests/src/test/java/org/sonarqube/gradle/Gradle9Test.java
@@ -49,14 +49,14 @@ public class Gradle9Test extends AbstractGradleIT {
       entry("sonar.java.source", "17"),
       entry("sonar.java.target", "17"),
       entry("sonar.java.test.binaries", "${parentBaseDir}/gradle-9-example/build/classes/java/test"),
-      entry("sonar.java.test.libraries", "${M2}/repository/org/apiguardian/apiguardian-api/1.1.2/apiguardian-api-1.1.2.jar," +
+      entry("sonar.java.test.libraries", "${parentBaseDir}/gradle-9-example/build/classes/java/main," +
         "${M2}/repository/org/jspecify/jspecify/1.0.0/jspecify-1.0.0.jar," +
-        "${M2}/repository/org/junit/jupiter/junit-jupiter-api/5.10.0/junit-jupiter-api-5.10.0.jar," +
-        "${M2}/repository/org/junit/jupiter/junit-jupiter-params/5.10.0/junit-jupiter-params-5.10.0.jar," +
         "${M2}/repository/org/junit/jupiter/junit-jupiter/5.10.0/junit-jupiter-5.10.0.jar," +
+        "${M2}/repository/org/junit/jupiter/junit-jupiter-params/5.10.0/junit-jupiter-params-5.10.0.jar," +
+        "${M2}/repository/org/junit/jupiter/junit-jupiter-api/5.10.0/junit-jupiter-api-5.10.0.jar," +
         "${M2}/repository/org/junit/platform/junit-platform-commons/1.10.0/junit-platform-commons-1.10.0.jar," +
         "${M2}/repository/org/opentest4j/opentest4j/1.3.0/opentest4j-1.3.0.jar," +
-        "${parentBaseDir}/gradle-9-example/build/classes/java/main"),
+        "${M2}/repository/org/apiguardian/apiguardian-api/1.1.2/apiguardian-api-1.1.2.jar"),
       entry("sonar.junit.reportPaths", "${parentBaseDir}/gradle-9-example/build/test-results/test"),
       entry("sonar.junit.reportsPath", "${parentBaseDir}/gradle-9-example/build/test-results/test"),
       entry("sonar.kotlin.gradleProjectRoot", "${parentBaseDir}/gradle-9-example"),
@@ -72,9 +72,9 @@ public class Gradle9Test extends AbstractGradleIT {
       entry("sonar.scanner.internal.dumpToFile", "<hidden>"),
       entry("sonar.scanner.os", "<hidden>"),
       entry("sonar.scanner.wasEngineCacheHit", "false"),
-      entry("sonar.sources", "${parentBaseDir}/gradle-9-example/build.gradle.kts," +
-        "${parentBaseDir}/gradle-9-example/settings.gradle.kts," +
-        "${parentBaseDir}/gradle-9-example/src/main/java"),
+      entry("sonar.sources", "${parentBaseDir}/gradle-9-example/src/main/java," +
+        "${parentBaseDir}/gradle-9-example/build.gradle.kts," +
+        "${parentBaseDir}/gradle-9-example/settings.gradle.kts"),
       entry("sonar.surefire.reportsPath", "${parentBaseDir}/gradle-9-example/build/test-results/test"),
       entry("sonar.tests", "${parentBaseDir}/gradle-9-example/src/test/java"),
       entry("sonar.working.directory", "${parentBaseDir}/gradle-9-example/build/sonar"));
--- a/src/main/java/org/sonarqube/gradle/AndroidUtils.java
+++ b/src/main/java/org/sonarqube/gradle/AndroidUtils.java
@@ -67,6 +67,7 @@ import org.gradle.util.GradleVersion;
 
 import static com.android.builder.model.Version.ANDROID_GRADLE_PLUGIN_VERSION;
 import static org.sonarqube.gradle.SonarUtils.appendSourcesProp;
+import static org.sonarqube.gradle.SonarUtils.exists;
 import static org.sonarqube.gradle.SonarUtils.nonEmptyOrNull;
 import static org.sonarqube.gradle.SonarUtils.setMainClasspathProps;
 import static org.sonarqube.gradle.SonarUtils.setTestClasspathProps;
@@ -345,7 +346,7 @@ class AndroidUtils {
       libraries.addAll(getLibraries((ApkVariant) variant));
     }
     if (javaCompile != null) {
-      libraries.addAll(javaCompile.getClasspath().filter(File::exists).getFiles());
+      libraries.addAll(exists(javaCompile.getClasspath()));
     }
 
     Collection<File> destinationDirs = (javaCompile != null)
@@ -364,7 +365,7 @@ class AndroidUtils {
       Method methodOnAndroidBefore30 = variant.getClass().getMethod("getCompileLibraries");
       return (Set<File>) methodOnAndroidBefore30.invoke(variant, (Object[]) null);
     } catch (NoSuchMethodException e) {
-      return variant.getCompileClasspath(null).getFiles();
+      return exists(variant.getCompileClasspath(null));
     } catch (IllegalAccessException | InvocationTargetException e) {
       throw new IllegalArgumentException("Unable to call getCompileLibraries", e);
     }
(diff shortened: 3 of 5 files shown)


integrationTests/src/test/resources/gradle-9-example/settings.gradle.kts
```diff
@@ -0,0 +1,17 @@
+rootProject.name = "gradle-9-example"
+
+pluginManagement {
+    resolutionStrategy {
+        eachPlugin {
+            if (requested.id.id == "org.sonarqube") {
+                useModule("org.sonarsource.scanner.gradle:sonarqube-gradle-plugin:${requested.version}")
+            }
+        }
```
dorian-burihabwa-sonarsource, line 9: Is this block necessary if the resolution strategy systematically starts by looking up in the local Maven repositories?

integrationTests/src/test/resources/gradle-9-example/.gitignore
```diff
@@ -0,0 +1,9 @@
+.gradle
+build/
+!gradle/wrapper/gradle-wrapper.jar
+!**/src/main/**/build/
+!**/src/test/**/build/
```
dorian-burihabwa-sonarsource: Do we need these patterns? The first one says that we should not ignore a file that is alreay committed anyway while the 2 other lines mention folders that are not supposed to exist

integrationTests/src/test/java/org/sonarqube/gradle/GradleTest.java
```diff
@@ -271,7 +297,8 @@ public void testProjectWithConfigurationCacheDoubleExecutionsShouldWork() throws
 
   @Test
   public void testKotlinMultiplatformProject() throws Exception {
-    Assume.assumeTrue("Tests only applies to version 6.8.3 or greater", getGradleVersion().isGreaterThanOrEqualTo("6.8.3"));
+    ignoreThisTestIfGradleVersionIsLessThan("6.8.3");
+    ignoreThisTestIfGradleVersionIsGreaterThanOrEqualTo("9.0.0");
```
dorian-burihabwa-sonarsource: It looks like a range method will eventually be needed if we cannot find a way to share these assertions somewhere. Let's not rush it and see if we need to add these before the end of the sprint

integrationTests/src/test/java/org/sonarqube/gradle/GradleTest.java
```diff
@@ -167,19 +187,19 @@ public void module_inclusion_duplicate_key() throws Exception {
...
-
-    assertThat(props).contains(entry(":toplevel1.:toplevel1:plugins.sonar.moduleKey", "com.mygroup:root_project:toplevel1:plugins"));
-    assertThat(props).contains(entry(":toplevel2.:toplevel2:plugins.sonar.moduleKey", "com.mygroup:root_project:toplevel2:plugins"));
+    assertThat(props)
+      .contains(entry(":toplevel1.sonar.moduleKey", "com.mygroup:root_project:toplevel1"))
+    .contains(entry(":toplevel2.sonar.moduleKey", "com.mygroup:root_project:toplevel2"))
+    .contains(entry(":toplevel1.sonar.modules", ":toplevel1:plugins"))
+    .contains(entry(":toplevel2.sonar.modules", ":toplevel2:plugins"))
+    .contains(entry(":toplevel1.:toplevel1:plugins.sonar.moduleKey", "com.mygroup:root_project:toplevel1:plugins"))
+    .contains(entry(":toplevel2.:toplevel2:plugins.sonar.moduleKey", "com.mygroup:root_project:toplevel2:plugins"));
```
dorian-burihabwa-sonarsource: Let's make the indentation consistent here
```suggestion
      .contains(entry(":toplevel1.sonar.moduleKey", "com.mygroup:root_project:toplevel1"))
      .contains(entry(":toplevel2.sonar.moduleKey", "com.mygroup:root_project:toplevel2"))
      .contains(entry(":toplevel1.sonar.modules", ":toplevel1:plugins"))
      .contains(entry(":toplevel2.sonar.modules", ":toplevel2:plugins"))
      .contains(entry(":toplevel1.:toplevel1:plugins.sonar.moduleKey", "com.mygroup:root_project:toplevel1:plugins"))
      .contains(entry(":toplevel2.:toplevel2:plugins.sonar.moduleKey", "com.mygroup:root_project:toplevel2:plugins"));
```

dorian-burihabwa-sonarsource (changes requested):
I added a couple of suggestions but I was thinking of another approach for the existing tests.

What about extracting the new test to a new test file such as `Gradle9Test`, and add the compatibility in `@Before` annotated method for each of `BootstrapTest`, `GradleTest` and `Gradle9Test`?

integrationTests/src/test/resources/gradle-9-example/settings.gradle.kts
alban-auzeill: I going to remove it to see if it's useful

integrationTests/src/test/resources/gradle-9-example/.gitignore
alban-auzeill: It's not useful for this project, so I will remove it. It was just generated by Intellij whe you create a new project. It's not the thing that I look at because the generation is usually not harmful.

integrationTests/src/test/java/org/sonarqube/gradle/AbstractGradleIT.java
```diff
@@ -67,6 +73,90 @@ protected static Semver getGradleVersion() {
...
+      "sonar.scanner.apiBaseUrl",
+      "sonar.scanner.app",
+      "sonar.scanner.appVersion",
+      "sonar.scanner.arch",
+      "sonar.scanner.internal.dumpToFile",
+      "sonar.scanner.os",
+      "sonar.scanner.wasEngineCacheHit"));
+    Map<String, String> replacementMap = new LinkedHashMap<>();
+    replacementMap.put("${parentBaseDir}", Paths.get(absoluteProjectBaseDir).getParent().toString());
+    replacementMap.put("${PWD}", System.getProperty("user.dir"));
```
dorian-burihabwa-sonarsource: Does PWD stand for project working directory? If so, I think it would be good to make it explicit to avoid confusion with the unix command `pwd`. WDYT?

integrationTests/src/test/java/org/sonarqube/gradle/AbstractGradleIT.java
```diff
@@ -67,6 +73,90 @@ protected static Semver getGradleVersion() {
...
+    Set<String> hiddenProperties = new HashSet<>(Arrays.asList(
+      // different on each machine, nothing to compare
+      "sonar.java.jdkHome", "sonar.scanner.arch", "sonar.scanner.os", "sonar.scanner.opts",
+      // path to a different temporary file for each run, e.g. junit17577551948278219342.tmp
+      "sonar.scanner.internal.dumpToFile",
+      // depends on the plugin and gradle version that run the test, e.g. 6.3-SNAPSHOT/Gradle 9.0.0
+      "sonar.scanner.appVersion"));
+    // All environment variables started with SONAR_SCANNER_* will pollute the test by adding sonar.scanner.* properties
+    // We are forced to filter them out
+    Set<String> testedSonarScannerProperties = new HashSet<>(Arrays.asList(
```
dorian-burihabwa-sonarsource: Can we rename this set to `sonarScannerPropertiesToTest` for clarity?

integrationTests/src/test/java/org/sonarqube/gradle/AbstractGradleIT.java
```diff
@@ -67,6 +73,90 @@ protected static Semver getGradleVersion() {
...
+    Assume.assumeTrue("Test is ignored for Gradle version " + gradleVersion + " which is greater than or equal to ", gradleVersion.isLowerThan(version));
+  }
+
+  protected static void ignoreThisTestIfGradleVersionIsNotBetween(String minVersionIncluded, String maxVersionExcluded) {
+    ignoreThisTestIfGradleVersionIsLessThan(minVersionIncluded);
+    ignoreThisTestIfGradleVersionIsGreaterThanOrEqualTo(maxVersionExcluded);
+  }
+
+  protected static Map<String, String> extractComparableProperties(Properties props) {
+    Map<String, String> result = new LinkedHashMap<>();
```
dorian-burihabwa-sonarsource: Should we move this variable closer to its usage (ie: just before we iterate over the properties)?

integrationTests/src/test/java/org/sonarqube/gradle/AbstractGradleIT.java
```diff
@@ -67,6 +73,90 @@ protected static Semver getGradleVersion() {
...
+        value = value.replace('\\', '/');
+        for (String replacementKey : replacementMap.keySet()) {
+          value = removeAbsolutePath("${M2}", replacementKey + "/.m2", value);
+        }
+        value = dependenciesInGradleCache.matcher(value).replaceAll(res ->
+        // Java 11 does not support res.group("groupId") this is why there is res.group(1)
+        "\\${M2}/repository/" + res.group(1).replace(".", "/") + "/${artifactId}/${version}/${artifactId}-${version}.jar");
+        result.put(key, value);
+      });
+    return result;
```
dorian-burihabwa-sonarsource: Nitpick: We can just generate the map from the stream itself to avoid jumping through to the `result` declaration.
```suggestion
    return props.stringPropertyNames().stream()
      .sorted()
      .filter(key -> !key.startsWith("sonar.scanner.") || testedSonarScannerProperties.contains(key))
      .map(key -> {
        String value = props.getProperty(key);
        if (hiddenProperties.contains(key)) {
          value = "<hidden>";
        }
        for (Map.Entry<String, String> entry : replacementMap.entrySet()) {
          value = removeAbsolutePath(entry.getKey(), entry.getValue(), value);
        }
        value = value.replace('\\', '/');
        for (String replacementKey : replacementMap.keySet()) {
          value = removeAbsolutePath("${M2}", replacementKey + "/.m2", value);
        }
        value = dependenciesInGradleCache.matcher(value).replaceAll(res ->
        // Java 11 does not support res.group("groupId") this is why there is res.group(1)
        "\\${M2}/repository/" + res.group(1).replace(".", "/") + "/${artifactId}/${version}/${artifactId}-${version}.jar");
        return new AbstractMap.SimpleEntry(key, value);
      }).collect(
        Collectors.toMap(
          Map.Entry<String, String>::getKey,
          Map.Entry<String, String>::getValue, 
          null,
          LinkedHashMap<String, String>::new
        )
      );
```

integrationTests/src/test/java/org/sonarqube/gradle/AbstractGradleIT.java
```diff
@@ -67,6 +73,90 @@ protected static Semver getGradleVersion() {
...
+  }
+
+  private static String removeAbsolutePath(String placeHolder, String absolutePath, String value) {
```
dorian-burihabwa-sonarsource: Nitpick: The method does not seem to be aware that parameters are path, can we rename this to `replaceAllPatterns(String placeHolder, String pattern, String value)` ?

integrationTests/pom.xml
```diff
@@ -49,6 +49,16 @@
...
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.8.1</version>
+        <configuration>
+          <release>11</release>
+        </configuration>
+      </plugin>
+    </plugins>
```
dorian-burihabwa-sonarsource, line 61: Should this change be documented (with a comment or in the README) to explain why we set a release target of 11?

dorian-burihabwa-sonarsource (changes requested):

integrationTests/pom.xml
alban-auzeill: This change only applies to the unit tests of `integrationTests`, it does not impact the build of the scanner. No need to change the README, there is already: `To build the plugin and run the tests, you will need Java 11.`. And the change is safe, we run the IT using `Dockerfile_java11` & `Dockerfile_java17`. So there is no reason not to benefit from Java 11 in IT.

integrationTests/src/test/java/org/sonarqube/gradle/AbstractGradleIT.java
alban-auzeill: I think PWD make sense because:
System.getProperty("user.dir") is the same as System.getenv("PWD") or Path.of(".").toRealPath()

integrationTests/src/test/java/org/sonarqube/gradle/AbstractGradleIT.java
alban-auzeill: I don't think it is easier to read if I replace:
```
    Map<String, String> result = new LinkedHashMap<>();
    ...
      result.put(key, value);
    ...
   return result;
```
by
```
    return ...
    ...
      return new AbstractMap.SimpleEntry(key, value);
    ...
    .collect(
        Collectors.toMap(
          Map.Entry<String, String>::getKey,
          Map.Entry<String, String>::getValue,
          null,
          LinkedHashMap<String, String>::new
        )
      );
```

dorian-burihabwa-sonarsource (approved):

PR #320, opened 2025-08-26
SCANGRADLE-251 Create GitHub workflow to bump development iteration

tomasz-tylenda-sonarsource:
[SCANGRADLE-251](https://sonarsource.atlassian.net/browse/SCANGRADLE-251)

It is not possible to test the workflow that does not exist on master yet, so this change has not been tested.

[SCANGRADLE-251]: https://sonarsource.atlassian.net/browse/SCANGRADLE-251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/PrepareNextIteration.yml
+++ b/.github/workflows/PrepareNextIteration.yml
@@ -24,6 +24,7 @@ jobs:
         run: |
           git config user.name "${GITHUB_ACTOR}"
           git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+          # Consider creating a unique branch name to avoid collisions when the action is run multiple times.
           git checkout -b gh-action/next-iteration
           pushd integrationTests
           mvn versions:set -DgenerateBackupPoms=false -DnewVersion="${{ inputs.nextVersion }}"


.github/workflows/PrepareNextIteration.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Checkout Sources
+        uses: actions/checkout@v4
+
+      - name: Update Version Number
+        run: |
+          git config user.name "${GITHUB_ACTOR}"
+          git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+          git checkout -b gh-action/next-iteration
+          pushd .
+          cd integrationTests
```
dorian-burihabwa-sonarsource: We should be able to do both operations in one, right?
```suggestion
          pushd integrationTests
```

.github/workflows/PrepareNextIteration.yml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+
+      - name: Checkout Sources
+        uses: actions/checkout@v4
+
+      - name: Update Version Number
+        run: |
+          git config user.name "${GITHUB_ACTOR}"
+          git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+          git checkout -b gh-action/next-iteration
```
dorian-burihabwa-sonarsource, line 28: Should be create with the version number to avoid collisions if the action is invoked multiple times?

.github/workflows/PrepareNextIteration.yml
```diff
@@ -0,0 +1,40 @@
...
+          popd
+          sed --in-place "s/version=.*$/version=${{ inputs.nextVersion }}/" gradle.properties
+          git commit -m "Prepare next development iteration ${{ inputs.nextVersion }}" -a
+          git push --set-upstream origin gh-action/next-iteration
+
+      - name: Create Pull Request
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          gh pr create -B master --title "Prepare next development iteration ${{ inputs.nextVersion }}" --body ""
```
dorian-burihabwa-sonarsource: Why don't we use long options consistently in this command?
```suggestion
          gh pr create --base master --title "Prepare next development iteration ${{ inputs.nextVersion }}" --body ""
```

.github/workflows/PrepareNextIteration.yml
tomasz-tylenda-sonarsource: I added a comment and will try to do this in the next PR. This action in other repos also also uses a fixed branch name and it hasn't been a problem in practice.

PR #349, opened 2025-09-15
SCANGRADLE-247 Class path resolution through a dedicated project based task

dorian-burihabwa-sonarsource:
[SCANGRADLE-247](https://sonarsource.atlassian.net/browse/SCANGRADLE-247)

[SCANGRADLE-247]: https://sonarsource.atlassian.net/browse/SCANGRADLE-247?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
+++ b/src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
@@ -113,10 +113,9 @@ public class SonarQubePlugin implements Plugin<Project> {
     topLevelProject.getAllprojects().forEach(target ->
       target.getTasks().register(SonarResolverTask.TASK_NAME, SonarResolverTask.class, task -> {
         if (target == topLevelProject) {
-          task.setProjectName("");
-        } else {
-          task.setProjectName(target.getName());
+          task.setTopLevelProject(true);
         }
+        task.setProjectName(target.getName());
         Configuration compileClasspath = target.getConfigurations().findByName(Constants.COMPILE_CLASSPATH);
         if (compileClasspath != null) {
           task.setCompileClasspath(compileClasspath);
--- a/src/main/java/org/sonarqube/gradle/SonarResolverTask.java
+++ b/src/main/java/org/sonarqube/gradle/SonarResolverTask.java
@@ -40,6 +40,7 @@ public abstract class SonarResolverTask extends DefaultTask {
   private static final Logger LOGGER = Logger.getLogger(SonarResolverTask.class.getName());
 
   private String projectName;
+  private boolean isTopLevelProject;
   private FileCollection compileClasspath;
   private FileCollection testCompileClasspath;
   private File outputDirectory;
@@ -51,6 +52,15 @@ public abstract class SonarResolverTask extends DefaultTask {
 
   public void setProjectName(String name) {
     this.projectName = name;
+  }
+
+  @Input
+  public boolean isTopLevelProject() {
+    return isTopLevelProject;
+  }
+
+  public void setTopLevelProject(boolean topLevelProject) {
+    this.isTopLevelProject = topLevelProject;
   }
 
   @InputFiles
@@ -80,7 +90,7 @@ public abstract class SonarResolverTask extends DefaultTask {
   @OutputFile
   public File getOutputFile() throws IOException {
     String filename;
-    if (projectName == null || projectName.isEmpty()) {
+    if (isTopLevelProject()) {
       filename = "properties";
     } else {
       filename = String.format("%s.properties", projectName);
@@ -98,9 +108,6 @@ public abstract class SonarResolverTask extends DefaultTask {
   @TaskAction
   void run() throws IOException {
     String displayName = getProjectName();
-    if (displayName.isEmpty()) {
-      displayName = "top-level project";
-    }
     if (LOGGER.isLoggable(Level.INFO)) {
       LOGGER.info("Resolving properties for " + displayName + ".");
     }


src/main/java/org/sonarqube/gradle/ResolutionSerializer.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  public static File write(File output, Map<String, List<File>> properties) throws IOException {
+    try (BufferedWriter writer = new BufferedWriter(new FileWriter(output, StandardCharsets.UTF_8))) {
+      for (Map.Entry<String, List<File>> entry : properties.entrySet()) {
+        String key = entry.getKey();
+        String absolutePaths = entry.getValue().stream()
+                .map(File::getAbsolutePath)
+                .collect(Collectors.joining(","));
+        writer.write(String.format("%s=%s%s", key, absolutePaths, System.lineSeparator()));
```
leonardo-pilastri-sonarsource:
```suggestion
            List<String> absolutePaths = entry.getValue().stream()
                .map(File::getAbsolutePath)
                .collect(Collectors.toList());
        String csvString = SonarUtils.joinAsCsv(absolutePaths);
        writer.write(String.format("%s=%s%s", key, csvString, System.lineSeparator()));
```

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -77,21 +81,61 @@ public void apply(Project project) {
...
   }
 
+  /**
+   * Register and configure a resolver task per (sub-)project.
+   * As the tasks are configured, we capture list of output files where the resolved properties will be written.
+   */
+  private static List<File> registerAndConfigureResolverTasks(Project topLevelProject, File sonarResolver) {
+    List<File> resolverFiles = new ArrayList<>();
+    topLevelProject.getAllprojects().forEach(target ->
+      target.getTasks().register("sonarResolver", SonarResolverTask.class, task -> {
```
leonardo-pilastri-sonarsource: Let's move this `"sonarResolver"` to a constant inside the SonarResolverTask class for consistency

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -77,21 +81,61 @@ public void apply(Project project) {
...
+        if (target == topLevelProject) {
+          task.setProjectName("");
```
leonardo-pilastri-sonarsource: Let's check if we can use the root project's name, if it helps with the other logic.
Maybe we can set a flag in the SonarResolverTask to distinguish between root and modules projects

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -153,6 +198,12 @@ private static Callable<Iterable<? extends Task>> getJavaCompileTasks(Project pr
       .collect(Collectors.toList());
   }
 
+  private static Callable<Iterable<? extends Task>> getClassPathResolverTask(Project project) {
+    return () -> project.getAllprojects().stream()
+            .map(p -> p.getTasks().getByName("sonarResolver"))
```
leonardo-pilastri-sonarsource: Let's use the constant here to make sure we depend on the right thing

src/main/java/org/sonarqube/gradle/SonarResolverTask.java
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  public void setOutputDirectory(File outputDirectory) {
+    this.outputDirectory = outputDirectory;
+  }
+
+  @OutputFile
+  public File getOutputFile() throws IOException {
+    String filename;
+    if (projectName == null || projectName.isEmpty()) {
```
leonardo-pilastri-sonarsource: Check if we can use without issues the root project name for the output filename

src/main/java/org/sonarqube/gradle/SonarResolverTask.java
```diff
@@ -0,0 +1,125 @@
...
+    if (!output.createNewFile()) {
+      throw new IOException("Could not create output file: " + output.getAbsolutePath());
+    }
+    return output;
+  }
+
+  @TaskAction
+  void run() throws IOException {
+    String displayName = getProjectName();
+    if (displayName.isEmpty()) {
```
leonardo-pilastri-sonarsource: If we introduce a flag let's use it here

src/main/java/org/sonarqube/gradle/SonarTask.java
```diff
@@ -194,6 +201,201 @@ public Provider<Map<String, String>> getProperties() {
...
+
+  /**
+   * Reads class path information produced as output of {@link SonarResolverTask}, regenerates related
+   * sonar.java.libraries and sonar.java.test.libraries and stores them into the result map.
+   */
+  private static void processResolverFile(File resolverFile, Map<String, String> result) {
+    LOGGER.info("Looking at file: {}", resolverFile);
+    String projectName = "";
+    String absolutePath = resolverFile.getAbsolutePath();
+    if (absolutePath.endsWith(".properties")) {
```
leonardo-pilastri-sonarsource, line 249: Let's check if using the root project's name in the earlier steps can simplify this logic and have it work the same here.

src/main/java/org/sonarqube/gradle/ResolutionSerializer.java
```diff
@@ -0,0 +1,73 @@
...
+        List<File> files = SonarUtils.splitAsCsv(librariesAsString).stream()
+                .map(File::new)
+                .collect(Collectors.toList());
+        properties.put(classPathKey, files);
+      }
+    }
+    return properties;
+  }
+
+  public static File write(File output, Map<String, List<File>> properties) throws IOException {
```
leonardo-pilastri-sonarsource, line 59: I was thinking that if it's not too much work, the write method could also take the project's name as input, and store it in the first line. Then when reading we would return a new class that contains both the project's name and the Map of properties with the lists of files. But only if this does not take too much time

src/main/java/org/sonarqube/gradle/SonarTask.java
leonardo-pilastri-sonarsource: If we decide to introduce a custom class that is deserialized from the resolver files we could use the project's name from there

src/main/java/org/sonarqube/gradle/SonarTask.java
```diff
@@ -194,6 +201,201 @@ public Provider<Map<String, String>> getProperties() {
...
+
+    if (LOGGER.isDebugEnabled()) {
+      LOGGER.debug("Resolved configured test class path as: {}", resolvedAsAString);
+    }
+
+    // Prepend sonar.java.binaries if it exists
+    String binariesPropertyKey = isTopLevelProject ?
+            "sonar.java.binaries" :
+            (":" + projectName + ".sonar.java.binaries");
+    String libraries = properties.getOrDefault(binariesPropertyKey, "");
```
leonardo-pilastri-sonarsource, line 368: I see that we do not do this for the main classpath, is this correct?

src/main/java/org/sonarqube/gradle/SonarTask.java
```diff
@@ -194,6 +201,201 @@ public Provider<Map<String, String>> getProperties() {
...
+
+    // Add existing test libraries if they exist
+    String propertyKey = isTopLevelProject ?
+            "sonar.java.test.libraries" :
+            (":" + projectName + ".sonar.java.test.libraries");
+
+    // Append resolved libraries
+    if (properties.containsKey(propertyKey)) {
```
leonardo-pilastri-sonarsource, line 376: What happens if the properties do not contain `:moduleA.sonar.java.test.libraries` yet? Do we discard all the collected properties? Can that happen?

leonardo-pilastri-sonarsource (changes requested):
I think some more tests could be nice especially for the serialization/deserialization part that was introduced, but let's just settle for the minimum coverage required for now.

src/main/java/org/sonarqube/gradle/SonarTask.java
dorian-burihabwa-sonarsource: That is correct. If we look at the properties dumped with previous versions of the analyzer, the binaries are always pre-pended to `sonar.java.test.libraries`.

src/main/java/org/sonarqube/gradle/SonarTask.java
dorian-burihabwa-sonarsource: That is a fair point, I have created a [ticket](https://sonarsource.atlassian.net/browse/SCANGRADLE-266) to tackle that here.

src/main/java/org/sonarqube/gradle/ResolutionSerializer.java
dorian-burihabwa-sonarsource: I suggest we tackle that with another [ticket](https://sonarsource.atlassian.net/browse/SCANGRADLE-266)

src/main/java/org/sonarqube/gradle/SonarTask.java
dorian-burihabwa-sonarsource: If the key is missing then we just skip. For the Gradle scanner, we should not try to make subproject inherit properties from other projects. From my tests, it looks consistent across projects.

PR #483, opened 2026-04-14
SCANGRADLE-315 Migrate Android specific code to AGP 9

aurelien-coet-sonarsource:


rombirli:
#### TODO
- [x] fix for agp7
- [x] Migrate the computation of JDK properties and binaries to the new API
- [x] Bug provider(?) in test libraries
- [x] Manifest 
- [x] Res
- [x] add agp9 to its matrix & migrate old projects to support agp9
- [x] understand & fix exception with logs in ConfigCache.java:57
- [x] fix rewrite-property-snapshot to allow gradle-version and agp-version overrides 
- [ ] Investigate issue with tests on agp9
- [ ] Coverage
- [ ] Cleaning history
- [x] Issues QG
- [x] Share and document python script to run/debug its

poseggaQM:
Our Sonar Pipeline is failing since we upgraded to AGP9, leaving us in a non-compliant state. If you can't fix this ASAP we will migrate to another vendor

PR #484, opened 2026-04-15
SCANGRADLE-275 Simplify path normalization

rombirli:


integrationTests/src/test/java/org/sonarqube/gradle/support/normalization/PathsNormalizer.java
```diff
@@ -23,187 +23,84 @@
...
-    for (Map.Entry<String, String> entry : PROPERTIES_REPLACEMENT.entrySet()) {
-      var key = entry.getKey();
-      var placeHolder = entry.getValue();
-      var value = snapshot.get(key);
-      Optional.ofNullable(value).ifPresent(v -> replacements.put(v, placeHolder));
+  private static String normalizeEntry(String property, String value, @Nullable String projectBaseDir) {
+    var paths = Arrays.stream(value.split(DELIMITER));
+    if (property.endsWith(BINARIES_SUFFIX) || property.endsWith(LIBRARIES_SUFFIX)) {
+      paths = paths
+        .map(PathsNormalizer::normalizeWindowsPath)
```
aurelien-coet-sonarsource, line 78: I see that `normalizeWindowsPath` is called on the argument of `takeOnlyFileName`, so I think the call to the method is redundant here.

aurelien-coet-sonarsource (approved):
Just a small comment, but otherwise LGTM !

PR #493, opened 2026-04-23
SCANGRADLE-315 Add support for AGP 9

aurelien-coet-sonarsource:
Related ticket: [SCANGRADLE-374](https://sonarsource.atlassian.net/browse/SCANGRADLE-374)

[SCANGRADLE-374]: https://sonarsource.atlassian.net/browse/SCANGRADLE-374?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/build.yml
```diff
@@ -111,11 +113,11 @@ jobs:
           - gradle_version: "8.0"
           - gradle_version: "7.6.4"
           - gradle_version: "7.5.1"
-            android_gradle_version: "7.1.0"
+            android_gradle_version: "7.4.0"
```
dorian-burihabwa-sonarsource, line 116: Why are changing AGP version here?

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -59,11 +59,7 @@
 public class SonarQubePlugin implements Plugin<Project> {
   private static final Logger LOGGER = Logging.getLogger(SonarQubePlugin.class);
 
-  private static ActionBroadcast<SonarProperties> addBroadcaster(Map<String, ActionBroadcast<SonarProperties>> actionBroadcastMap, Project project) {
-    return actionBroadcastMap.computeIfAbsent(project.getPath(), s -> new ActionBroadcast<>());
-  }
-
-  private static boolean addTaskByName(Project p, String name, List<Task> allCompileTasks) {
+  static boolean addTaskByName(Project p, String name, List<Task> allCompileTasks) {
```
dorian-burihabwa-sonarsource: Is the visibility change intentional?

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -72,79 +68,91 @@ private static boolean addTaskByName(Project p, String name, List<Task> allCompi
...
-      TaskContainer tasks = project.getTasks();
-      tasks.register(SonarExtension.SONAR_DEPRECATED_TASK_NAME, SonarTask.class, task -> {
-        task.setDescription("Analyzes " + project + " and its subprojects with Sonar. This task is deprecated. Use 'sonar' instead.");
-        task.setGroup(JavaBasePlugin.VERIFICATION_GROUP);
-        task.setResolverFiles(resolverFiles);
-        task.setBuildSonar(project.getLayout().getBuildDirectory().dir("sonar"));
-        configureTask(task, project, actionBroadcastMap);
-      });
+  private static ActionBroadcast<SonarProperties> addBroadcaster(Map<String, ActionBroadcast<SonarProperties>> actionBroadcastMap, Project project) {
+    return actionBroadcastMap.computeIfAbsent(project.getPath(), s -> new ActionBroadcast<>());
```
dorian-burihabwa-sonarsource:
```suggestion
    return actionBroadcastMap.computeIfAbsent(project.getPath(), ignored -> new ActionBroadcast<>());
```

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -72,79 +68,91 @@ private static boolean addTaskByName(Project p, String name, List<Task> allCompi
...
+      resolverTask.setProjectName(SonarUtils.constructPrefixedProjectName(project.getPath()));
+      if (project == topLevelProject) {
+        resolverTask.setTopLevelProject(true);
+      }
+      resolverTask.setCompileClasspath(project.provider(() -> querySourceSet(project, SourceSet.MAIN_SOURCE_SET_NAME)));
+      resolverTask.setTestCompileClasspath(project.provider(() -> querySourceSet(project, SourceSet.TEST_SOURCE_SET_NAME)));
+      if (!isAndroidProject(project)) {
+        resolverTask.setMainLibraries(project.provider(() -> project.files(SonarUtils.getRuntimeJars())));
+        resolverTask.setTestLibraries(project.provider(() -> project.files(SonarUtils.getRuntimeJars())));
+      } else if (!AndroidConfig.usesAndroidGradlePlugin9()) {
```
dorian-burihabwa-sonarsource: From the name of the method, it seems to be strictly about using AGP9 but from the block below, it seems to be about being strictly lower than AGP9. Can we rename the test method to reflect that?
```suggestion
      } else if (!AndroidConfig.usesAndroidGradlePlugin9OrGreater()) {
```

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -72,79 +68,91 @@ private static boolean addTaskByName(Project p, String name, List<Task> allCompi
...
+              resolverTask.setMainLibraries(project.provider(androidConfig::getMainLibraries));
+              resolverTask.setTestLibraries(project.provider(androidConfig::getTestLibraries));
+              resolverTask.setAndroidSources(project.provider(androidConfig::getAndroidSources));
+              resolverTask.setAndroidTests(project.provider(androidConfig::getAndroidTests));
+              resolverTask.mustRunAfter(androidConfig.getTasks());
+            });
+          })
+        );
+      }
+    } catch (NoClassDefFoundError e) {
```
dorian-burihabwa-sonarsource:
```suggestion
    } catch (NoClassDefFoundError ignored) {
```

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -72,79 +68,91 @@ private static boolean addTaskByName(Project p, String name, List<Task> allCompi
...
+      // The Android plugin is not available in the project, so we do not configure Android.
```
dorian-burihabwa-sonarsource, line 144: Should we log at trace level to show that we do not configure Android properties for this specific subproject?

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -72,79 +68,91 @@ private static boolean addTaskByName(Project p, String name, List<Task> allCompi
...
+      resolverTask.setCompileClasspath(project.provider(() -> querySourceSet(project, SourceSet.MAIN_SOURCE_SET_NAME)));
+      resolverTask.setTestCompileClasspath(project.provider(() -> querySourceSet(project, SourceSet.TEST_SOURCE_SET_NAME)));
+      if (!isAndroidProject(project)) {
+        resolverTask.setMainLibraries(project.provider(() -> project.files(SonarUtils.getRuntimeJars())));
+        resolverTask.setTestLibraries(project.provider(() -> project.files(SonarUtils.getRuntimeJars())));
+      } else if (!AndroidConfig.usesAndroidGradlePlugin9()) {
+        resolverTask.setMainLibraries(project.provider(() -> LegacyAndroidConfig.findMainLibraries(project)));
+        resolverTask.setTestLibraries(project.provider(() -> LegacyAndroidConfig.findTestLibraries(project)));
+        resolverTask.mustRunAfter(getAndroidTasks(project));
+      }
```
dorian-burihabwa-sonarsource: Would it make sense to move this bit of logic into `configureAndroid` below? Following the flow the resolver task registration and configuration, the libraries for android subprojects could be configured in in of 2 places. It might help to get keep the android library logic into the same location.

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -72,79 +68,91 @@ private static boolean addTaskByName(Project p, String name, List<Task> allCompi
...
-      })
-    );
-    return resolverFiles;
+  /**
+   * Configure Android specific properties and classpath information for a project if it uses the Android Gradle plugin.
+   */
+  private static void configureAndroid(Project project, Map<String, AndroidConfig> androidConfigMap, TaskProvider<SonarResolverTask> resolverTaskProvider) {
+    try {
+      if (AndroidConfig.usesAndroidGradlePlugin9()) {
+        SonarUtils.ANDROID_PLUGIN_IDS.forEach(pluginId ->
```
dorian-burihabwa-sonarsource, line 123: Is it likely to have multiple Android plugin one the same project? If so, the properties from last plugin in the list would be used as the source of truth. Correct?

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -215,7 +216,7 @@ private static Callable<Iterable<? extends Task>> getJavaCompileTasks(Project pr
       .collect(Collectors.toList());
   }
 
-  private static Callable<Iterable<? extends Task>> getClassPathResolverTask(Project project) {
+  private static Callable<Iterable<? extends Task>> getClassPathResolverTasks(Project project) {
```
dorian-burihabwa-sonarsource, line 211: 👍🏿

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -72,79 +68,91 @@ private static boolean addTaskByName(Project p, String name, List<Task> allCompi
...
+        SonarUtils.ANDROID_PLUGIN_IDS.forEach(pluginId ->
+          project.getPlugins().withId(pluginId, plugin -> {
+            AndroidConfig androidConfig = AndroidConfig.of(project);
+            androidConfigMap.put(project.getPath(), androidConfig);
+            resolverTaskProvider.configure(resolverTask -> {
+              resolverTask.setMainLibraries(project.provider(androidConfig::getMainLibraries));
+              resolverTask.setTestLibraries(project.provider(androidConfig::getTestLibraries));
+              resolverTask.setAndroidSources(project.provider(androidConfig::getAndroidSources));
+              resolverTask.setAndroidTests(project.provider(androidConfig::getAndroidTests));
+              resolverTask.mustRunAfter(androidConfig.getTasks());
```
dorian-burihabwa-sonarsource, line 132: We changed the scheduling constraints from having the both all android tasks and all resolver tasks running before the sonar task to having all android tasks running before all resolver tasks running before the sonar task.
I am assuming the change is intentional to have the resolver task work with resolved configurations from the android tasks. Is that corrrect?
Whether it is this or something different, it might be worth documenting why we have this new scheduling constraint.

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
```diff
@@ -263,12 +254,51 @@ private static Callable<Iterable<? extends Task>> getAndroidTasks(Project projec
...
           addTaskByName(p, testTaskPrefix + "UnitTest", allTasks);
         }
         return allTasks;
       })
       .flatMap(List::stream)
       .collect(Collectors.toList());
   }
+
+  @Nullable
+  static String getConfiguredAndroidVariant(Project p) {
```
dorian-burihabwa-sonarsource: I see that the method existed in this file before but does it make sense to move it to SonarUtils?

dorian-burihabwa-sonarsource (commented):
Pausing review after review of SonarQubePlugin. Will continue later

src/main/java/org/sonarqube/gradle/SonarUtils.java
```diff
@@ -60,7 +68,7 @@ public class SonarUtils {
   /**
    * Find test files given the path by looking for the keyword "test", for example:
    * - script/test/run.sh
-   *          ^^^^
+   * ^^^^
```
dorian-burihabwa-sonarsource: That should probably not have been reformatted

src/main/java/org/sonarqube/gradle/SonarUtils.java
```diff
@@ -77,12 +85,11 @@ private SonarUtils() {
     // Utility class
   }
 
+  /**
+   * Check if a Gradle project is an Android project.
```
dorian-burihabwa-sonarsource:
```suggestion
   * Check if a Gradle project is an Android project by looking up known plugins.
```

src/main/java/org/sonarqube/gradle/SonarUtils.java
```diff
@@ -331,6 +338,7 @@ public static Set<Path> extractReportPaths(Map<String, Object> properties) {
 
   /**
    * Computes the absolute paths for the report paths extracted from the properties.
+   *
```
dorian-burihabwa-sonarsource: Can be reverted
```suggestion
```

src/main/java/org/sonarqube/gradle/ProjectProperties.java
```diff
@@ -42,42 +42,75 @@
...
+   * @param compileClasspath     resolved compile classpath as absolute paths
    * @param testCompileClasspath resolved test compile classpath as absolute paths
-   * @param mainLibraries filtered main libraries for analysis
-   * @param testLibraries filtered test libraries for analysis
+   * @param mainLibraries        filtered main libraries for analysis
+   * @param testLibraries        filtered test libraries for analysis
    */
-  public ProjectProperties(String projectName, Boolean isRootProject, List<String> compileClasspath, List<String> testCompileClasspath,
-                           List<String> mainLibraries, List<String> testLibraries) {
+  @SuppressWarnings("java:S107")
```
dorian-burihabwa-sonarsource, line 95: It might make sense to create a ticket to come back here later and redesign this class with a builder pattern.

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
aurelien-coet-sonarsource: Yes, this method is now also used in `AndroidConfig`. Maybe it would make sense to move it to `SonarUtils`.

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
aurelien-coet-sonarsource: We could, however this means that we would also log this information for projects that have nothing at all to do with Android.

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
aurelien-coet-sonarsource: I believe it is not possible to apply different Android plugins to the same project, so at most one can be picked up here.

src/main/java/org/sonarqube/gradle/SonarQubePlugin.java
aurelien-coet-sonarsource: Yes, the resolver tasks need to run after Android tasks to get their resolved configurations, and the Sonar task needs to run after all resolver tasks to get the resolved classpaths/Android sources, so transitively the Sonar task necessarily runs after all Android tasks. I'll document this 👍

src/main/java/org/sonarqube/gradle/SonarTask.java
```diff
@@ -316,133 +221,31 @@ static void processResolverFile(File resolverFile, Map<String, String> result) {
...
-    }
-
-    // remove xml report if directory do not exist
-    List<PropertyInfo> xmlReportProperties = parsePropertiesWithNames(properties, Set.of(SonarProperty.JACOCO_XML_REPORT_PATHS));
-    for (PropertyInfo prop : xmlReportProperties) {
-      properties.computeIfPresent(prop.fullName, (k, commaList) -> {
-        var filtered = filterPaths(commaList, Files::exists, userDefinedKeys.contains(k));
-        return filtered.isEmpty() ? null : filtered;
-      });
+  static void resolveSources(ProjectProperties projectProperties, @Nullable Collection<File> sources, Map<String, String> properties, boolean isTest) {
```
dorian-burihabwa-sonarsource: It looks like this method is specific to android sources. Should we rename it to reflect that?
```suggestion
  static void resolveAndroidSources(ProjectProperties projectProperties, @Nullable Collection<File> sources, Map<String, String> properties, boolean isTest) {
```

.github/workflows/build.yml
aurelien-coet-sonarsource: I believe we changed it while iterating, when we were trying to support AGP <9 and >=9 in the same place, and forgot to put it back to its initial value. I'll switch it back.

src/main/java/org/sonarqube/gradle/SonarTask.java
```diff
@@ -589,17 +571,75 @@ public void useLoggerLevel(LogLevel logLevel) {
...
+    } catch (Exception e) {
+      throw new AnalysisException(e);
+    }
+  }
+
   /**
-   * @return The {@link LogOutput} object to use during Scanner execution. All logged messages from the Scanner will
-   * pass through this object. If needed, a custom implementation can be used to handle logged output, such as printing
-   * {@link LogLevel#INFO}-level log output when Gradle is only configured at the {@link LogLevel#LIFECYCLE} level.
+   * Finish the configuration of `sonar.sources`, `sonar.java.libraries` and `sonar.java.test.libraries` by resolving the Android sources and class paths that
```
dorian-burihabwa-sonarsource: Should that comment mention `sonar.tests` as well?

src/main/java/org/sonarqube/gradle/SonarTask.java
dorian-burihabwa-sonarsource, line 1: Looks good overall but the moving around of riles makes it a little difficult to see if there was any change in some of the methods that were simply moved up or down the file.

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
```diff
@@ -286,10 +292,10 @@ private static void computeScanAllProperties(Project project, Map<String, Object
...
     String existingValue = (String) properties.getOrDefault(sourcePropertyToUpdate, "");
     Set<Path> existingSources = existingValue.isBlank() ? Collections.emptySet() : SonarUtils.splitAsCsv(existingValue)
-      .stream()
-      .filter(Predicate.not(String::isBlank))
-      .map(Paths::get)
-      .collect(Collectors.toSet());
+                                                                                   .stream()
+                                                                                   .filter(Predicate.not(String::isBlank))
+                                                                                   .map(Paths::get)
+                                                                                   .collect(Collectors.toSet());
```
dorian-burihabwa-sonarsource: Is this intentional?

dorian-burihabwa-sonarsource (commented):
Tasks reviewed. Pausing for now. Will review AndroidConfig and LegacyAndroidConfig before the end of the day

src/main/java/org/sonarqube/gradle/ProjectProperties.java
aurelien-coet-sonarsource: Yes, I'll create a ticket for this.

src/main/java/org/sonarqube/gradle/SonarTask.java
aurelien-coet-sonarsource: Yes, sorry about that. It was a bad call to do a clean up and move methods around as part of the PR.

src/main/java/org/sonarqube/gradle/SonarPropertyComputer.java
aurelien-coet-sonarsource: No, this was probably a result of auto-formatting the file. I'll fix it.

.github/workflows/build.yml
aurelien-coet-sonarsource: It looks like there are some DSL elements that are needed for projects to be built with AGP 9 that only appeared in AGP 7.4+, which is why we cannot support both AGP 7.1 and 9 on the same projects in the integration tests, and why we had to bump this version number.

dorian-burihabwa-sonarsource: Then it's all good with me

src/main/java/org/sonarqube/gradle/LegacyAndroidConfig.java
```diff
@@ -66,18 +66,18 @@
...
 
 import static com.android.builder.model.Version.ANDROID_GRADLE_PLUGIN_VERSION;
-import static org.sonarqube.gradle.SonarQubePlugin.getConfiguredAndroidVariant;
+import static org.sonarqube.gradle.SonarUtils.getConfiguredAndroidVariant;
 import static org.sonarqube.gradle.SonarUtils.appendProps;
 import static org.sonarqube.gradle.SonarUtils.appendSourcesProp;
 
 /**
  * Only access this class when running on an Android application
  */
```
dorian-burihabwa-sonarsource, line 77: Can we have  the comment explain what should be logically grouped in this file rather than AndroidConfig? Does this class only contain constructs that are pre-AGP9? Or is the delineation more subtle than that?

src/main/java/org/sonarqube/gradle/AndroidConfig.java
```diff
@@ -0,0 +1,429 @@
...
+public class AndroidConfig {
+
+  private static final Logger LOGGER = Logging.getLogger(AndroidConfig.class);
+
+  private final Project project;
+  private final AndroidComponentsExtension<?, ?, ?> androidComponentsExtension;
+  private final List<Variant> variants;
+  private Variant selectedVariant;
+
+  @SuppressWarnings({"java:S1612", "java:S1602"})
```
dorian-burihabwa-sonarsource: I see we cannot easily get rid of S1612 but w can probably get rid of S1602, here

src/main/java/org/sonarqube/gradle/AndroidConfig.java
```diff
@@ -0,0 +1,429 @@
...
+    AndroidComponentsExtension<?, ?, ?> androidComponentsExtension = project.getExtensions().getByType(AndroidComponentsExtension.class);
+    var androidConfig = new AndroidConfig(project, androidComponentsExtension);
+    androidComponentsExtension.onVariants(androidComponentsExtension.selector().all(), variant -> {
+      androidConfig.variants.add(variant);
+    });
+    return androidConfig;
+  }
+
+  public static boolean usesAndroidGradlePlugin9OrGreater() {
+    return Version.of(ANDROID_GRADLE_PLUGIN_VERSION).compareTo(Version.of("9.0.0")) >= 0;
```
dorian-burihabwa-sonarsource, line 96: Is the value of `ANDROID_GRADLE_PLUGIN_VERSION` expected to change depending on the Gradle runtime? Because the method does not take any parameter, it might be worth explaining how and when the value is expected to change. If my project does not contain any Android construct, is the method safe to use? How is it expected to behave?

src/main/java/org/sonarqube/gradle/AndroidConfig.java
```diff
@@ -0,0 +1,429 @@
...
+        .getAsFile();
+      return Collections.singletonList(defaultJavaPath);
+    });
+  }
+
+  private static String getCompileTaskName(Component component) {
+    return "compile" + SonarUtils.capitalize(component.getName()) + "JavaWithJavac";
+  }
+
+  private AndroidConfig(Project project, AndroidComponentsExtension<?, ?, ?> androidComponentsExtension) {
```
dorian-burihabwa-sonarsource: I speak I, I think it would be nice to have the constructor declared closer to the object field. I initially wrote a comment about having an explicit private constructor but that was because I expected it to show up closer to the top of the class

src/main/java/org/sonarqube/gradle/AndroidConfig.java
```diff
@@ -0,0 +1,429 @@
...
+      selectedVariant = configuredVariant.get();
+    } else {
+      // Find the variant that is the target for Android tests in the project. If no variant has Android tests, return the first we find.
+      selectedVariant = variants.stream()
+        .filter(v -> v.getNestedComponents().stream().anyMatch(AndroidTest.class::isInstance))
+        .findFirst()
+        .orElse(variants.get(0));
+    }
+
+    return selectedVariant;
```
dorian-burihabwa-sonarsource: We could early return here
```suggestion
      }
      return configuredVariant.get();
    }
    // Find the variant that is the target for Android tests in the project. If no variant has Android tests, return the first we find.
    return variants.stream()
      .filter(v -> v.getNestedComponents().stream().anyMatch(AndroidTest.class::isInstance))
      .findFirst()
      .orElse(variants.get(0));
```

src/main/java/org/sonarqube/gradle/AndroidConfig.java
```diff
@@ -0,0 +1,429 @@
...
+   * Get the test components for the selected Android variant.
+   */
+  private List<Component> getTestComponents() {
+    List<Component> testComponents = new ArrayList<>();
+    for (Component component : getVariant().getNestedComponents()) {
+      if (component instanceof TestComponent) {
+        testComponents.add(component);
+      }
+    }
+    return testComponents;
```
dorian-burihabwa-sonarsource: Could we return this as an immutable list? Looking at the call sites, I don't see any reason to return a modifiable list to the caller. In a lot of places of this project, we pass a lot of collections around that are eventually modified by different components. I think setting a stricter mutability contract would help us to diagnose issues and be more aligned with stricter isolation and immutability constraint that Gradle keeps bringing

src/main/java/org/sonarqube/gradle/AndroidConfig.java
```diff
@@ -0,0 +1,429 @@
...
+      sourceFiles.from(sources.getByName("c").getAll());
+      sourceFiles.from(sources.getByName("cpp").getAll());
+
+      Method getRs = sources.getClass().getMethod("getRenderscript");
+      Object rs = getRs.invoke(sources);
+      if (rs != null) {
+        Method getAll = rs.getClass().getMethod("getAll");
+        sourceFiles.from(getAll.invoke(rs));
+      }
+    } catch (Exception ignored) {
```
dorian-burihabwa-sonarsource: We could be a little more explicit in the type of issues we accept to ignore
```suggestion
    } catch (InvocationTargetException | NoSuchMethodException | IllegalAccessException ignored) {) {
```

src/main/java/org/sonarqube/gradle/AndroidConfig.java
```diff
@@ -0,0 +1,429 @@
...
+  }
+
+  /**
+   * Get the source directories for the selected Android variant's tests.
+   */
+  public FileCollection getAndroidTests() {
+    FileCollection tests = project.files();
+    List<Component> testComponents = getTestComponents();
+    if (testComponents.isEmpty()) {
+      var sourceSet = project.getExtensions().getByType(CommonExtension.class).getSourceSets().findByName("test");
```
dorian-burihabwa-sonarsource, line 187: Can we document why expect some projects not to have test components? Can we default to extracting the source set from `CommonExtension` in all other cases?

src/main/java/org/sonarqube/gradle/AndroidConfig.java
```diff
@@ -0,0 +1,429 @@
...
+   */
+  public FileCollection getAndroidSources() {
+    return getSources(getVariant());
+  }
+
+  /**
+   * Get the source directories for the selected Android variant's tests.
+   */
+  public FileCollection getAndroidTests() {
+    FileCollection tests = project.files();
```
dorian-burihabwa-sonarsource: We can probably initialize this closer to where we use in the `else` block below

src/main/java/org/sonarqube/gradle/AndroidConfig.java
```diff
@@ -0,0 +1,429 @@
...
+    boolean testTaskAdded = false;
+    for (Component component : getTestComponents()) {
+      testTaskAdded |= addTaskByName(project, getCompileTaskName(component), tasks);
+    }
+    // The compilation of unit tests or Android tests already depends on the main compilation task, so it is only necessary to add it if no test compilation tasks were found.
+    if (!testTaskAdded) {
+      addTaskByName(project, getCompileTaskName(variant), tasks);
+    }
+
+    addTaskByName(project, "test" + SonarUtils.capitalize(variant.getName()) + "UnitTest", tasks);
```
dorian-burihabwa-sonarsource, line 223: Can we add a comment to explain why extract this specific test task outside of the loop?

src/main/java/org/sonarqube/gradle/AndroidConfig.java
```diff
@@ -0,0 +1,429 @@
...
+  /**
+   * Compute the Android lint report path for a given Android variant and populate properties with it.
+   */
+  private void configureLintReports(Map<String, Object> properties) {
+    project.getTasks().withType(AndroidLintTask.class).stream()
+      .filter(task -> task.getXmlReportOutputFile().isPresent())
+      .filter(task -> task.getVariantName().equals(getVariant().getName()))
+      .map(task -> task.getXmlReportOutputFile().get().getAsFile())
+      .findFirst()
+      .ifPresent(output -> properties.put(SonarProperty.ANDROID_LINT_REPORT_PATHS, output));
```
dorian-burihabwa-sonarsource, line 332: We can probably extract the getVariant computation ahead of the streaming.
```suggestion
String variantName = getVariant().getName();
    project.getTasks().withType(AndroidLintTask.class).stream()
      .filter(task -> task.getXmlReportOutputFile().isPresent())
      .filter(task -> task.getVariantName().equals(variantName))
      .map(task -> task.getXmlReportOutputFile().get().getAsFile())
      .findFirst()
      .ifPresent(output -> properties.put(SonarProperty.ANDROID_LINT_REPORT_PATHS, output));
```

src/main/java/org/sonarqube/gradle/AndroidConfig.java
```diff
@@ -0,0 +1,429 @@
...
+
+  public static boolean usesAndroidGradlePlugin9OrGreater() {
+    return Version.of(ANDROID_GRADLE_PLUGIN_VERSION).compareTo(Version.of("9.0.0")) >= 0;
+  }
+
+  private static int getMinSdk(Variant variant) {
+    return variant.getMinSdk().getApiLevel();
+  }
+
+  private static Provider<List<File>> getCompiledClasses(Project project, Component component) {
```
dorian-burihabwa-sonarsource: Is there a reason why the signature return a collection rather than a single file?

src/main/java/org/sonarqube/gradle/AndroidConfig.java
```diff
@@ -0,0 +1,429 @@
...
+   */
+  private Provider<List<RegularFile>> getManifestsProvider(Component component) {
+    Sources sources = component.getSources();
+    try {
+      // Try the modern Sources API (Added in AGP 8.3.1)
+      Method getManifestsMethod = sources.getClass().getMethod("getManifests");
+      Object manifestFiles = getManifestsMethod.invoke(sources);
+      Method getAllMethod = manifestFiles.getClass().getMethod("getAll");
+      return (Provider<List<RegularFile>>) getAllMethod.invoke(manifestFiles);
+    } catch (Exception e) {
```
dorian-burihabwa-sonarsource: Let's be explicit about the set of errors we expect
```suggestion
    } catch (InvocationTargetException | NoSuchMethodException  | IllegalAccessException ignored) {
```

dorian-burihabwa-sonarsource (commented):
Done with the review of AndroidConfig and Legacy AndroidConfig. I will resume with the review of the integration tests next time

integrationTests/pom.xml
```diff
@@ -15,8 +15,8 @@
 
   <properties>
     <!-- Default version when running ITs on dev machines. Value overridden in CI -->
-    <gradle.version>7.5.1</gradle.version>
-    <androidGradle.version>7.1.0</androidGradle.version>
+    <gradle.version>8.4</gradle.version>
+    <androidGradle.version>8.3.0</androidGradle.version>
```
dorian-burihabwa-sonarsource, line 19: Any reason why these values should be the new baseline and not some others?

src/main/java/org/sonarqube/gradle/LegacyAndroidConfig.java
aurelien-coet-sonarsource: Yes, I'll add information about this in the docstring. The class only contains constructs for AGP versions <9, whereas the `AndroidConfig` class is focused only on AGP >=9.

src/main/java/org/sonarqube/gradle/AndroidConfig.java
aurelien-coet-sonarsource: Yes, it makes sense. I'll move it up before the static methods.

src/main/java/org/sonarqube/gradle/AndroidConfig.java
aurelien-coet-sonarsource: Yes, this value is expected to change depending on the version of the Android plugin used at runtime for a project. The method doesn't take any parameter because this constant is "set in stone" and only depends on the version of AGP used at runtime. It is indeed unsafe to use in a project that doesn't use AGP.
You're right to call this out, I'll document this information in the method's docstring.

src/main/java/org/sonarqube/gradle/AndroidConfig.java
aurelien-coet-sonarsource: As discussed in person, I'll mark these two issues as FPs so that we can address them at some point in the analyzer.

src/main/java/org/sonarqube/gradle/AndroidConfig.java
aurelien-coet-sonarsource: No, I'll fix this.

integrationTests/src/test/java/org/sonarqube/gradle/AndroidTest.java
```diff
@@ -46,29 +47,10 @@ public static void beforeAll() {
     assumeTrue(getAndroidGradleVersion().isGreaterThan("7.0.0"));
   }
 
-  @Test
-  public void testSonarTaskHasNoDependencies() throws Exception {
```
dorian-burihabwa-sonarsource, line 50: Why was this test removed? Could we simply update the list of dependencies?

integrationTests/src/test/java/org/sonarqube/gradle/AndroidTest.java
```diff
@@ -46,29 +47,10 @@ public static void beforeAll() {
...
-        ":module-flavor1-androidTest-only:compileDebugJavaWithJavac SKIPPED",
-        ":module-plain-java:compileTestJava SKIPPED",
-        ":module_android_feature:compileFlavor1DebugJavaWithJavac SKIPPED")
-      );
-  }
-
   @Test
   public void gradle9AndroidParallelExample() throws Exception {
     ignoreThisTestIfGradleVersionIsLessThan("9.0.0");
+    Assume.assumeFalse(getAndroidGradleVersion().isGreaterThanOrEqualTo("9.0.0"));
```
dorian-burihabwa-sonarsource: Should this stay in the code?

integrationTests/src/test/java/org/sonarqube/gradle/support/normalization/PathsNormalizer.java
```diff
@@ -49,6 +49,10 @@ public class PathsNormalizer {
     "main", "R.jar", "classes.jar", "classes"
   );
 
+  private static final List<String> IGNORED_PATHS_SUBSTRINGS = List.of(
+    "kotlin-stdlib", "annotations-", "core-runtime-", "lifecycle-common-"
```
dorian-burihabwa-sonarsource, line 60: What is the rational behind removing libraries that contain these substrings? In which environment did we observe a difference? Is the difference likely to occur in production for some users?

dorian-burihabwa-sonarsource (approved):

integrationTests/pom.xml
aurelien-coet-sonarsource: These were picked as a middle ground which allows us to support AGP 7 and 9 at the same time with minimal use of reflection in the code.

integrationTests/src/test/java/org/sonarqube/gradle/AndroidTest.java
aurelien-coet-sonarsource: Yes, with Gradle 9+ and AGP 8+ the test fails because of issues with logs. I added a `FIXME` so that we revisit this issue in a later iteration.

integrationTests/src/test/java/org/sonarqube/gradle/support/normalization/PathsNormalizer.java
aurelien-coet-sonarsource: I've documented this.

integrationTests/src/test/java/org/sonarqube/gradle/support/normalization/PathsNormalizer.java
```diff
@@ -46,7 +46,18 @@ public class PathsNormalizer {
...
+    "classes",
+    "R.jar" // In AGP 9+, this JAR was removed from the classpath of Android applications. To keep supporting both AGP <9 and >=9, we ignore it in the tests.
+  );
+
+  /**
+   * The substrings in this list correspond to the names of JARs that vary across combinations of Gradle and AGP versions. To support several of these combinations on the same
+   * projects in the integration tests, we ignore JARs that contain these patterns.
+   */
+  private static final List<String> ANDROID_IGNORED_PATHS_SUBSTRINGS = List.of(
+    "kotlin-stdlib", "annotations-", "core-runtime-", "lifecycle-common-"
```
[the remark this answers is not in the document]
aurelien-coet-sonarsource: This is intended: the `support-annotations` library was deprecated and is no longer included in projects using more recent AGP versions. To support both older and newer AGP versions in the tests, we ignore it.

PR #500, opened 2026-05-07
SCANGRADLE-408 Fix CI : vulnerabilities in AGP 8.1.1

rombirli:


rombirli:
API broken, will keep it for later

PR #502, opened 2026-05-12
SCANGRADLE-353 Build the scanner with Java 21

aurelien-coet-sonarsource:


.github/workflows/build.yml
```diff
@@ -99,74 +99,60 @@ jobs:
...
+            java_version: "11"
           - gradle_version: "5.6.4"
+            java_version: "11"
     name: >
-      QA J11 - Gradle ${{ matrix.item.gradle_version }} 
+      QA J${{ matrix.item.java_version }} - Gradle ${{ matrix.item.gradle_version }}
       ${{ matrix.item.android_gradle_version && format(' Android {0}', matrix.item.android_gradle_version) || '' }}
+    runs-on: 'github-ubuntu-latest-l'
+    permissions:
+      id-token: write
```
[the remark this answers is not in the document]
aurelien-coet-sonarsource: Yes, this is intended.

.github/workflows/build.yml
```diff
@@ -99,74 +99,60 @@ jobs:
...
             android_gradle_version: "7.4.0"
+            java_version: "11"
           - gradle_version: "6.8.3"
+            java_version: "11"
           - gradle_version: "5.6.4"
+            java_version: "11"
     name: >
-      QA J11 - Gradle ${{ matrix.item.gradle_version }} 
+      QA J${{ matrix.item.java_version }} - Gradle ${{ matrix.item.gradle_version }}
       ${{ matrix.item.android_gradle_version && format(' Android {0}', matrix.item.android_gradle_version) || '' }}
```
dorian-burihabwa-sonarsource: Opportunity: We could try and be more explcit bout what `Android` refers to in this specific name

```suggestion
      ${{ matrix.item.android_gradle_version && format(' AGP {0}', matrix.item.android_gradle_version) || '' }}
```

integrationTests/src/test/java/org/sonarqube/gradle/BootstrapTest.java
dorian-burihabwa-sonarsource, line 60: Probably should be part of another PR and ticket: I feel like the heuristic we are using here is off w.r.t with the purpose of the test. The bootstrapping mechanism was introduced as part of SQ 10.6. In theory the guard could accept more specific conditions for testing
```java
var sonarRuntimeVersion = System.getProperty("sonar.runtimeVersion", "LATEST_RELEASE");

    if ((sonarRuntimeVersion.isGreaterOrEqualToThan("LATEST_RELEASE[26.1]") && getJavaVersion() == 21) ||
        (sonarRuntimeVersion.isGreaterOrEqualToThan("LATEST_RELEASE[10.6]") && sonarRuntimeVersion.isLowerThan("LATEST_RELEASE[26.1]") && getJavaVersion() == 17)
    ) {
      ORCHESTRATOR = OrchestratorRule.builderEnv()
        .setSonarVersion(sonarRuntimeVersion)
        .useDefaultAdminCredentialsForBuilds(true)
        .addPlugin(MavenLocation.of("org.sonarsource.java", "sonar-java-plugin", "LATEST_RELEASE"))
        .addPlugin(FileLocation.of("../property-dump-plugin/target/property-dump-plugin-1-SNAPSHOT.jar"))
        .build();
    } else {
      ORCHESTRATOR = null;
    }
```
The reason why I think it would be a little tricky to handle right now is that the java analyzer version would need to be consistent as well. WDYT about creating a ticket for that?

gradle/verification-metadata.xml
```diff
@@ -2361,20 +2363,14 @@
             <sha256 value="83408107930e1683a9b7647da7ab75eb83d80fa470c1641b79b5aecbbef1b783" origin="Verified"/>
          </artifact>
       </component>
```
dorian-burihabwa-sonarsource: Can we get rid of this? Or have it in a future ticket

dorian-burihabwa-sonarsource (approved):
LGTM. I left a few comments for potential improvements

integrationTests/src/test/java/org/sonarqube/gradle/BootstrapTest.java
aurelien-coet-sonarsource: I agree, it would make sense to be less strict and test more configurations here. I'll create a ticket for this.

gradle/verification-metadata.xml
aurelien-coet-sonarsource: Yes, my bad, I missed this remaining metadata for 0.8.7.

PR #510, opened 2026-06-02
SCANGRADLE-402 Redesign the ProjectProperties class with the builder pattern

NoemieBenard:
Replace the simple constructor of the ProjectProperties class with a Builder pattern.


aurelien-coet-sonarsource (approved):
LGTM !

PR #523, opened 2026-06-30
SCANGRADLE-429 Deduplicate paths in sonar.sources and sonar.tests with android KMP project

Seppli11:
Part of 
 

Please be aware that we are not actively looking for feature contributions. The truth is that it's extremely difficult for someone outside SonarSource to comply with our roadmap and expectations. Therefore, we typically only accept minor cosmetic changes and typo fixes. If you would like to see a new feature, please create a new thread in the forum ["Suggest new features"](https://community.sonarsource.com/c/suggestions/features).

With that in mind, if you would like to submit a code contribution, make sure that you adhere to the following guidelines and all tests are passing:

- [ ] Please explain your motives to contribute this change: what problem you are trying to fix, what improvement you are trying to make
- [ ] Use the following formatting style: [SonarSource/sonar-developer-toolset](https://github.com/SonarSource/sonar-developer-toolset#code-style)
- [ ] Provide a unit test for any code you changed
- [ ] If there is a [JIRA](http://jira.sonarsource.com/browse/SONARGRADL) ticket available, please make your commits and pull request start with the ticket ID (SONARGRADL-XXXX)

We will try to give you feedback on your contribution as quickly as possible.

Thank You!
The SonarSource Team

----
## Summary by Gitar

- **Deduplication logic:**
  - Added `mergePathCsvWithoutDuplicates` to `SonarTask` to handle path normalization and prevent redundant entries in `sonar.sources` and `sonar.tests`.
- **Integration testing:**
  - Added new Android KMP integration test projects `kmp-android-double-indexing` and `kmp-android-double-indexing-new-dsl` to verify path handling in multi-platform builds.
- **Refactoring:**
  - Updated `SonarResolverTask` to improve null safety when retrieving absolute paths from `FileCollection` providers.
- **Testing:**
  - Added comprehensive unit tests in `SonarTaskTest` to verify path deduplication and normalization across various scenarios.


src/main/java/org/sonarqube/gradle/SonarTask.java
```diff
@@ -242,12 +244,38 @@ static void resolveAndroidSources(ProjectProperties projectProperties, @Nullable
...
+    Map<String, String> pathsByNormalizedPath = new LinkedHashMap<>();
+    List<String> paths = new ArrayList<>();
+    paths.addAll(SonarUtils.splitAsCsv(existingPaths));
+    paths.addAll(SonarUtils.splitAsCsv(resolvedPaths));
+
+    for (String path : paths) {
+      if (path.isBlank()) {
+        continue;
+      }
+      pathsByNormalizedPath.putIfAbsent(deduplicationKey(path), path);
```
dorian-burihabwa-sonarsource, line 266: Due to the order of insertion, the assumption is that existing paths are always prioritized over resolved paths. Is this intended? If so, can make that explicit in the method documentation?

src/main/java/org/sonarqube/gradle/SonarTask.java
```diff
@@ -242,12 +244,38 @@ static void resolveAndroidSources(ProjectProperties projectProperties, @Nullable
...
   }
 
+  private static String mergePathCsvWithoutDuplicates(String existingPaths, String resolvedPaths) {
+    Map<String, String> pathsByNormalizedPath = new LinkedHashMap<>();
+    List<String> paths = new ArrayList<>();
+    paths.addAll(SonarUtils.splitAsCsv(existingPaths));
+    paths.addAll(SonarUtils.splitAsCsv(resolvedPaths));
+
+    for (String path : paths) {
+      if (path.isBlank()) {
```
dorian-burihabwa-sonarsource, line 263: In addition to deduplicating paths, this method removes blank entries. Can we make that clear in the method documentation or tests?

src/main/java/org/sonarqube/gradle/SonarTask.java
```diff
@@ -242,12 +244,38 @@ static void resolveAndroidSources(ProjectProperties projectProperties, @Nullable
...
+  }
+
+  private static String deduplicationKey(String path) {
+    if (containsWildcard(path)) {
+      return path;
+    }
+
+    try {
+      return Path.of(path).toAbsolutePath().normalize().toString();
+    } catch (InvalidPathException e) {
```
dorian-burihabwa-sonarsource, line 279: Should we log (at DEBUG or TRACE level) when a path is invalid? My understanding is that this code is only called at execution time (ie: when the analysis is about to start and files submitted for analysis should be available). Maybe we can make the exception buble up to `mergePathCsvWithoutDuplicates` and log there?

src/main/java/org/sonarqube/gradle/SonarResolverTask.java
```diff
@@ -157,7 +159,10 @@ public void setSkipProject(Provider<Boolean> skipProject) {
    * Returns the absolute paths of the files in the given FileCollection provider.
    * If the provider is null or the collection is empty, returns an empty list.
    */
-  private static List<String> getAbsolutePaths(Provider<FileCollection> fileCollection) {
+  private static List<String> getAbsolutePaths(@Nullable Provider<FileCollection> fileCollection) {
+    if (fileCollection == null) {
```
dorian-burihabwa-sonarsource: Do we know in which case we might get a nullable provider? I don't see any changes in the tests or elsewhere related in the code

dorian-burihabwa-sonarsource (commented):
Pausing review for now, I stopped at the changes on main code and sill need to look at the functional tests

src/main/java/org/sonarqube/gradle/SonarTask.java
Seppli11: :+1: Makes sense.

Just to clarify, is the removal of blank entries safe? I assumed, but maybe there is a purpose having blank entries

EDIT: I've added a test case for this

src/main/java/org/sonarqube/gradle/SonarTask.java
Seppli11: A debug statement would make sense. But I think we want to keep the behavior and keep the path even in the case where it is invalid, since the path might have been supplied by the user. Additionally, if we filter these paths, that would mean that invalid paths get filtered but only if it is an android project. This feels odd and inconsistent. 

Maybe this is an argument to move the logic further up the processing and make it apply more globally to the `sonar.sources` and `sonar.tests` properties, instead of just in android projects.

EDIT: I've added a debug statement in the `deduplicationKey` method and kept the exception there

src/main/java/org/sonarqube/gradle/SonarResolverTask.java
Seppli11: There was a case where the providers weren't set and were therefor still `null`. If memory serves me right, in one unit test it blew up because of this.

src/main/java/org/sonarqube/gradle/SonarTask.java
Seppli11: I added a test case for it

src/main/java/org/sonarqube/gradle/SonarTask.java
dorian-burihabwa-sonarsource: Yes, I think we can assume removing blank spaces is OK

src/test/java/org/sonarqube/gradle/SonarTaskTest.java
```diff
@@ -244,6 +248,91 @@ void resolveAndroidSources_combinesWithExistingPropertyToAddSources(@TempDir Fil
...
+    File sourceDirectory = new File(tempDir, "src/main/java");
+    sourceDirectory.mkdirs();
+
+    SonarTask.resolveAndroidSources(projectProperties, List.of(sourceDirectory), properties, false);
+
+    assertThat(SonarUtils.splitAsCsv(properties.get("sonar.sources")))
+      .containsExactly(existingPath, sourceDirectory.getAbsolutePath());
+  }
+
+  static Stream<Arguments> notNormalizedExistingAndroidSourcePaths() {
```
dorian-burihabwa-sonarsource: Nitpick: can we name this argument provider more clearly?

```suggestion
  static Stream<Arguments> nonNormalizablePaths() {
```

src/test/projects/kmp-android-double-indexing/gradle.properties
```diff
@@ -0,0 +1,4 @@
+org.gradle.jvmargs=-Xmx2g -Dfile.encoding=UTF-8
+kotlin.code.style=official
+android.builtInKotlin=false
+android.newDsl=false
```
dorian-burihabwa-sonarsource, line 4: Any reason why we need to set `android.newDsl` to false? We used to have  bug in the previous version of the scanner but we should be able to deal with newer versions of AGP now

src/test/projects/kmp-android-double-indexing/gradle.properties
```diff
@@ -0,0 +1,4 @@
+org.gradle.jvmargs=-Xmx2g -Dfile.encoding=UTF-8
```
dorian-burihabwa-sonarsource: Can we remove this line?
```suggestion
```

src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
```diff
@@ -800,6 +801,42 @@ class FunctionalTests extends Specification {
...
+    when:
+    def result = GradleRunner.create()
+      .withProjectDir(kmpAndroidProjectDir.toFile())
+      .withGradleVersion("9.5.1")
+      .forwardOutput()
+      .withArguments(
+        'sonar',
+        '--info',
+        '-Dsonar.scanner.internal.dumpToFile=' + outFile.toAbsolutePath(),
+        '-DsonarPluginClasspath=' + pluginClasspath()
```
dorian-burihabwa-sonarsource, line 823: Is there a more idiomatic way of passing this information?

dorian-burihabwa-sonarsource: I see th ecomment above but it would be really useful to point to a documented issue on the Gradle project if `withPluginClasspath()` cannot be used

src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
```diff
@@ -800,6 +801,42 @@ class FunctionalTests extends Specification {
 
   }
 
+  @Requires({ System.getenv("JAVA_HOME") != null && System.getenv("ANDROID_HOME") != null })
+  def "KMP Android resolver sources do not duplicate KMP source directories"() {
```
dorian-burihabwa-sonarsource: Should the title of this test read?
```suggestion
  def "Android resolved sources do not duplicate KMP source directories"() {
```

dorian-burihabwa-sonarsource (approved):
The change looks good to me but I am wondering if we could:
1. Make the project under test simpler
2. Link to the Gradle Testkit issue forcing us to pass the plugin classpath

src/test/projects/kmp-android-double-indexing/gradle.properties
Seppli11: Removed it :+1:

PR #535, opened 2026-08-20
SCANGRADLE-448 Update sonar-scanner-java-library to 4.2.1.1698

alban-auzeill:
[SCANGRADLE-448](https://sonarsource.atlassian.net/browse/SCANGRADLE-448)
----
## Summary by Gitar

- **Dependencies:**
    - Updated `sonar-scanner` library to `4.2.0.1683` using the shaded fat jar version
    - Updated verification metadata and updated `Gson` imports to use shaded internal paths

[SCANGRADLE-448]: https://sonarsource.atlassian.net/browse/SCANGRADLE-448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gradle/libs.versions.toml
henryju:
```suggestion
sonar-scanner-library = { group = "org.sonarsource.scanner.lib", name = "sonar-scanner-java-library-shaded", version.ref = "sonar-scanner" }
```

integrationTests/src/test/java/org/sonarqube/gradle/snapshot/SnapshotIO.java
```diff
@@ -19,9 +19,9 @@
  */
 package org.sonarqube.gradle.snapshot;
 
-import com.google.gson.Gson;
-import com.google.gson.GsonBuilder;
-import com.google.gson.reflect.TypeToken;
+import org.sonarsource.scanner.lib.internal.shaded.com.google.gson.Gson;
```
henryju, line 22: Please add an explicit dependency on gson for the integration tests, to not rely on shaded dependencies.

src/main/java/org/sonarqube/gradle/ResolutionSerializer.java
```diff
@@ -19,8 +19,8 @@
  */
 package org.sonarqube.gradle;
 
-import com.google.gson.Gson;
-import com.google.gson.GsonBuilder;
+import org.sonarsource.scanner.lib.internal.shaded.com.google.gson.Gson;
```
henryju, line 22: Same

henryju (changes requested):
Maybe also take the opportunity to drop the `gradle/verification-metadata.xml` as we discussed with the squad?

src/main/java/org/sonarqube/gradle/ResolutionSerializer.java
henryju: Sorry, this one is production code, so possibly better to use the shaded dependency than adding another transitive dependency, that could also conflict with other Gradle plugins.

henryju (approved):

PR #536, opened 2026-08-25
[DO NOT MERGE] Test bouncycastle bcpkix

alban-auzeill:
----
## Summary by Gitar

- **Dependency updates:**
    - Updated `sonar-scanner` version to `4.2.0.1675` in `libs.versions.toml`
    - Added verification metadata for `org.bouncycastle` and `sonar-scanner-java-library` dependencies


alban-auzeill:
We will not use this solution. We assume it works because bcpkix-jdk18on 1.85 depends on bcprov-jdk18on 1.85, not 1.85.2, but this is not a guarantee for future releases.

PR #539, opened 2026-08-27
SCANGRADLE-445 Update kotlin-gradle-plugin and remove as compileOnly dependency

Seppli11:


gradle/verification-metadata.xml
```diff
@@ -2250,12 +2250,14 @@
             <sha256 value="87721cbaa65a3c97d8b1ba9d207840f164c9fe38759fc9ea10ffe26565f8d3e9" origin="Verified"/>
          </artifact>
       </component>
-      <component group="org.jetbrains.kotlin" name="kotlin-android-extensions" version="1.8.21">
-         <artifact name="kotlin-android-extensions-1.8.21.jar">
-            <sha256 value="7468557042b350db853c1684cc2e24d37b5e4a1e09220abca93a50c7ba91c08f" origin="Verified"/>
+      <component group="org.jetbrains.kotlin" name="fus-statistics-gradle-plugin" version="2.4.0">
+         <artifact name="fus-statistics-gradle-plugin-2.4.0-gradle813.jar">
+            <sha256 value="032d24532d648e42302348a2c80a5046dc9f4520c64bbd203e7e040df6c8691f" origin="Verified"/>
```
dorian-burihabwa-sonarsource, line 2255: Why do we need 2.4.0 when we have 2.4.10 a few lines below?

dorian-burihabwa-sonarsource: The same question repeats for all entries related 2.4.0 below?

dorian-burihabwa-sonarsource (changes requested):
I have reviewed the entries 2.4.10 but do we need to keep the entries for 2.4.0?

gradle/verification-metadata.xml
Seppli11: My bad. I first upgrade to 2.4.0 and then to 2.4.10 but didn't remove the old entries.

dorian-burihabwa-sonarsource (approved):

PR #540, opened 2026-08-27
SCANGRADLE-446 Adds m2-sha256-verifier.py script

Seppli11:
Adds the `m2-sha256-verifer.py` script that allows for easy verifying of the `verification-metadata.xml` file


scripts/m2-sha256-verifier.py
```diff
@@ -0,0 +1,227 @@
...
+from dataclasses import dataclass, field
+from pathlib import Path
+
+REPOSITORIES = {
+    "Maven Central": "https://repo1.maven.org/maven2",
+    "Gradle Plugin Portal": "https://plugins.gradle.org/m2",
+    "Google Maven": "https://dl.google.com/dl/android/maven2",
+}
+METADATA_PATH = (
+    Path(__file__).resolve().parent.parent / "gradle" / "verification-metadata.xml"
```
dorian-burihabwa-sonarsource, line 19: Is that portable?
```suggestion
    os.path.join(Path(__file__).resolve().parent.parent, "gradle", "verification-metadata.xml")
```

scripts/m2-sha256-verifier.py
```diff
@@ -0,0 +1,227 @@
...
+    unverifiable_count = 0
+
+    for verification in verifications:
+        checksum = verification.checksum
+        print(f"\n{checksum.coordinate} ({checksum.artifact_name})")
+        print(f"  Reported checksum: {checksum.expected_sha256}")
+        has_match = False
+        has_mismatch = False
+
+        for repository in REPOSITORIES:
```
dorian-burihabwa-sonarsource, line 148: We can probably exit the loop when a match or a mismatch has been detected, correct?

scripts/m2-sha256-verifier.py
```diff
@@ -0,0 +1,227 @@
...
+        f"{unverifiable_count} unverifiable"
+    )
+    return mismatched_count == 0 and unverifiable_count == 0
+
+
+def fetch_repository_results(
+    checksums: list[ArtifactChecksum],
+) -> list[ArtifactVerification]:
+    verifications = [ArtifactVerification(checksum) for checksum in checksums]
+    max_workers = min(12, len(checksums) * len(REPOSITORIES))
```
dorian-burihabwa-sonarsource, line 187: Why is 12 hardcoded?

scripts/m2-sha256-verifier.py
```diff
@@ -0,0 +1,227 @@
...
+            print("  Result: UNVERIFIABLE (no repository returned a checksum)")
+
+    print(
+        f"\nSummary: {verified_count} verified, {mismatched_count} mismatched, "
+        f"{unverifiable_count} unverifiable"
+    )
+    return mismatched_count == 0 and unverifiable_count == 0
+
+
+def fetch_repository_results(
```
dorian-burihabwa-sonarsource, line 183: I think that by default, we might want to save the repository results in-between runs rather than hammering these repositories on every run.

scripts/m2-sha256-verifier.py
```diff
@@ -0,0 +1,227 @@
...
+                print(
+                    f"  [OK] {result.repository:<20} {result.sha256} ({result.source})"
+                )
+                has_match = True
+            else:
+                print(
+                    f"  [!!] {result.repository:<20} {result.sha256} "
+                    f"(expected {checksum.expected_sha256})"
+                )
+                has_mismatch = True
```
dorian-burihabwa-sonarsource, line 164: With these conditionals, we are testing the same that the expected checksums match for all available reposiotry entries. Do you think it would make sense to document against which repository we matched and against which we did not?

README.md
```diff
@@ -89,9 +89,7 @@ And delete in the `verification-metadata.xml` the versions that we don't use any
...
 
-When you update a dependency’s checksum in the gradle/verification-metadata.xml file, you validate the change by comparing the sha256 value from Artifactory with the one listed on another package repositories like maven central. 
-First, identify the dependency updated in the file and copy its sha256 checksum. 
-Next, search for the dependency on another package repositories. For instance on [Maven Central](https://central.sonatype.com), you can use the query `checksum:new-dependency-sha256` to find a specific dependency.
+When you update a dependency’s checksum in the gradle/verification-metadata.xml file, you validate the change by comparing the sha256 value from Artifactory with the one listed on another package repositories like maven central. This can be done by using the `scripts/m2-sha256-verifier.py` script that compares the sha256 value present int he `gradle/verification-metadata.xml` file with the artifact on Maven Central, Gradle Plugin Portal or Google Maven. If all entries match the reported sha256 value, the `Generated by Gradle` can be replaced with `Verified`.
```
dorian-burihabwa-sonarsource, line 92:
```suggestion
When you update a dependency’s checksum in the gradle/verification-metadata.xml file, you validate the change by comparing the sha256 value from Artifactory with the one listed on another package repository like maven central. This can be done by using the `scripts/m2-sha256-verifier.py` script that compares the sha256 value present int he `gradle/verification-metadata.xml` file with the artifact on Maven Central, Gradle Plugin Portal or Google Maven. If all entries match the reported sha256 value, the `Generated by Gradle` can be replaced with `Verified`. 
```

dorian-burihabwa-sonarsource (approved):
Nothing blocking here but there a couple of opportunities for improvement

scripts/m2-sha256-verifier.py
dorian-burihabwa-sonarsource: That's on me, I was not familar with [Pathlib](https://docs.python.org/3/library/pathlib.html#basic-use)

Seppli11:
For now, I'll let this PR stay open until the weekly on the 02.09.2026 where we'll discuss if we want to remove the `verification-metadata.xml` file entirely.

Seppli11:
Given that we'll remove the verification-metadata.xml file entirely, this PR is no longer needed

PR #542, opened 2026-09-01
SCANGRADLE-449 sonarResolver fails on composite builds: "Cannot use mustRunAfter to reference tasks from another build"

alban-auzeill:
----
## Summary by Gitar

- **Bug fix:**
    - Fixed `sonarResolver` failing on composite builds by ignoring classpath producer tasks from included builds.

### DIFF
--- a/src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
+++ b/src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
@@ -1133,6 +1133,51 @@ class FunctionalTests extends Specification {
     resolverOnlyProperties.compileClasspath.contains(jvmJarPath)
   }
 
+  def "sonarResolver ignores classpath producers coming from an included build"() {
+    given:
+    settingsFile << """
+        rootProject.name = 'root'
+        includeBuild('included-lib') {
+            dependencySubstitution {
+                substitute module('com.example:included-lib') using project(':')
+            }
+        }
+        """
+    buildFile << """
+        plugins {
+            id 'org.sonarqube'
+            id 'java-library'
+        }
+
+        dependencies { implementation 'com.example:included-lib:1.0' }
+        """
+    writeFile(projectDir.resolve("included-lib/settings.gradle"), "rootProject.name = 'included-lib'")
+    writeFile(projectDir.resolve("included-lib/build.gradle"), """
+        plugins { id 'java-library' }
+        group = 'com.example'
+        version = '1.0'
+        """)
+    writeFile(projectDir.resolve("included-lib/src/main/java/example/Library.java"), "package example;\npublic class Library {}")
+    writeFile(projectDir.resolve("src/main/java/example/Consumer.java"), "package example;\npublic class Consumer {}")
+
+    when:
+    def result = GradleRunner.create()
+      .withProjectDir(projectDir.toFile())
+      .withGradleVersion("9.5.1")
+      .forwardOutput()
+      .withPluginClasspath()
+      .withArguments(':compileJava', ':sonarResolver')
+      .build()
+
+    then:
+    result.task(":sonarResolver").getOutcome() == SUCCESS
+    !result.output.contains("Failed to resolve classpath producer tasks")
+    !result.output.contains("Failed to resolve file collection input")
+    def resolverProperties = new JsonSlurper().parse(projectDir.resolve("build/sonar-resolver/properties").toFile())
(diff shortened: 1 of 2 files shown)


alban-auzeill:
> Code Review: Fixes `sonarResolver` failing on composite builds by removing the `mustRunAfter` constraint to included-build classpath producers. However, this removes the only scheduling guarantee between `sonarResolver` and those producers, allowing `sonarResolver` to execute before the included build's output is ready. Since `SonarUtils.exists()` silently drops missing classpath entries, the serialized `compileClasspath` can be incomplete with no warning, degrading analysis instead of failing loudly. The test should be split into separate build and resolve invocations to verify deterministic ordering, and missing classpath entries from other builds should be logged as a warning so reduced coverage is diagnosable.

  Rejected: splitting the test into two invocations. The claim was that compileClasspath.contains(included-lib/.../classes/java/main) "only holds because Gradle happens to run the requested entry tasks in order — there is no declared ordering." There is. SonarQubePlugin.java:108 already declares
  resolverTask.mustRunAfter(compileJava, compileTestJava), and root :compileJava has a real dependsOn edge on the substituted included-build artifact. So the chain :included-lib:compileJava → :compileJava → (mustRunAfter) → :sonarResolver is fully declared.                                       
                                                                                                                                                                                                                                                                                                        
  I verified this rather than asserting it: I reversed the arguments to ':sonarResolver', ':compileJava' and the observed execution order was still [:included-lib:compileJava, :compileJava, :sonarResolver]. Splitting into two invocations would have made the test strictly weaker — it would stop  
  exercising ordering within a single task graph. Instead I kept the single invocation, made the reversed order permanent, and added an explicit assertion (FunctionalTests.groovy:1178) so a future ordering regression fails the test rather than passing silently.

src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
```diff
@@ -1133,6 +1133,51 @@ class FunctionalTests extends Specification {
...
+        group = 'com.example'
+        version = '1.0'
+        """)
+    writeFile(projectDir.resolve("included-lib/src/main/java/example/Library.java"), "package example;\npublic class Library {}")
+    writeFile(projectDir.resolve("src/main/java/example/Consumer.java"), "package example;\npublic class Consumer {}")
+
+    when:
+    def result = GradleRunner.create()
+      .withProjectDir(projectDir.toFile())
+      .withGradleVersion("9.5.1")
```
dorian-burihabwa-sonarsource, line 1166: Any reason why we pick this specific version?

src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
```diff
@@ -1133,6 +1133,51 @@ class FunctionalTests extends Specification {
...
+      .forwardOutput()
+      .withPluginClasspath()
+      .withArguments(':compileJava', ':sonarResolver')
+      .build()
+
+    then:
+    result.task(":sonarResolver").getOutcome() == SUCCESS
```
dorian-burihabwa-sonarsource, line 1173: We trigger `compileJava`, should we also check it ended successfully?

src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
```diff
@@ -1133,6 +1133,51 @@ class FunctionalTests extends Specification {
...
+    !result.output.contains("Failed to resolve classpath producer tasks")
```
dorian-burihabwa-sonarsource, line 1174: Are these strings Gradle version specific?

dorian-burihabwa-sonarsource (approved):

src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
alban-auzeill: This ensures a deterministic assertion of the error message.

src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
alban-auzeill: It fails before if the code does not compile

src/test/groovy/org/sonarqube/gradle/FunctionalTests.groovy
alban-auzeill: I think so

PR #547, opened 2026-09-21
SCANGRADLE-455 Update kotlin-gradle-plugin to the latest version

alban-auzeill:


src/test/projects/kmp-android-double-indexing-new-dsl/build.gradle.kts
```diff
@@ -23,7 +23,7 @@ extensions.configure(org.sonarqube.gradle.SonarExtension::class.java) {
 }
 
 plugins {
-    id("org.jetbrains.kotlin.multiplatform") version "2.3.0"
+    id("org.jetbrains.kotlin.multiplatform") version "2.4.20"
```
dorian-burihabwa-sonarsource: Do we need to upgrade the dependencies of this test project? Dependencies of test projects should be ignored during analysis

src/test/projects/kmp-android-double-indexing/build.gradle.kts
```diff
@@ -5,7 +5,7 @@ buildscript {
         gradlePluginPortal()
     }
     dependencies {
-        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:2.3.0")
+        classpath("org.jetbrains.kotlin:kotlin-gradle-plugin:2.4.20")
```
dorian-burihabwa-sonarsource: Same question about test projects?

dorian-burihabwa-sonarsource (approved):
LGTM. The tests seem to be passing even when the test projects are not updated but the upgrade does not seem to hurt here.

